PR #23, opened 2016-01-20
Add documentation from ringpop-node

jwolski:
Now that ringpop-common is open source we can begin to move the docs/ and tools/ (that are cross-platform) out of ringpop-node and into ringpop-common. Here are all the docs from the ringpop-node repo.

@uber/ringpop


thanodnl:
lgtm

dansimau:
lgtm

PR #25, opened 2016-01-21
Fix README to point to ringpop-go testpop path

dansimau:
@uber/ringpop


jwolski:
This path assumes that you built testpop with `make testpop`. Is that right? If so, I'd suggest an additional note in the README about that.

dansimau:
Good point. I actually forgot to change a part of the README. Will fix.

jwolski:
wewt. lgtm

PR #27, opened 2016-01-21
Log test debug info to stdout

dansimau:
These messages are extra information for the test run and not actual
error output from the test hardness itself. The rest of the test data is
output on stdout (e.g. tap data) so do that here as well.


jwolski:
I'm OK with these changes if it unblocks your path. Btw, what's your mission here?

Generally speaking, I think these integration tests need to have better error handling.

CorgiMan:
lgtm

dansimau:
@jwolski Totally agree.

In the mean time, I'm trying to work around that. The goal is that TravisCI only outputs the test failures.

PR #28, opened 2016-01-22
Integration test improvements

dansimau:
OK last one guys, I promise.

This PR includes two changes:

1. Make integration tests exit with non-zero code on failures

  Enough said.

2. Change behaviour of tap-filter, for the ringpop-go integration tests

  This is a bit of a change of direction. It turns out it is actually more sane just to provide a `-v` switch, where by default we print only failures and `-v` will print successes as well. And do all this on stdout. This matches the behaviour of, say, `go test -v`.

Thanks!

@uber/ringpop


benfleis:
lgtm

CorgiMan:
lgtm

jwolski:
lgtm3

PR #29, opened 2016-02-09
tap-filter take two (fix integration tests)

dansimau:
The previous tap-filter was using a node library which meant debug lines
were silently discarded (because they are not part of the tap format).

The new tap-filter behaves the same way as the previous one (printing
only errors), except that in addition to printing the tap data it also
prints any output that was buffered in the failing test.

This allows us to see useful debug output on Travis, as well as
information about which test failing assertions are in.


test/tap-filter
```diff
@@ -0,0 +1,76 @@
...
+	case $line in
+		\#*)
+			# Beginning of new test...
+
+			# If there was an error during the last test, print the error
+			# before resetting
+			if $test_error; then
+				echo >&2
+				echo -e "\033[1;31m${test_name}\033[0m" >&2
+				echo "$buffer" >&2
```
Motiejus: Will the buffer be printed if the last test failed?

dansimau: Yes... I think so... that is the intention....

CorgiMan:
lgtm Dannnn

CorgiMan:
Let's fix them tests

test/tap-filter
```diff
@@ -0,0 +1,76 @@
...
+
+while IFS= read -r line; do
+	case $line in
+		\#*)
+			# Beginning of new test...
+
+			# If there was an error during the last test, print the error
+			# before resetting
+			if $test_error; then
+				echo >&2
```
Motiejus: Nitpick: I think it would be easier to read, and harder to miss, if we have an explicit echo-like function which prints to stdout. E.g.

```bash
_log() {
  echo "$@" >&2
}
```
... and us that instead of `>&2` after every echo.

I will not push hard for it, just think about it.

test/tap-filter
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
Motiejus: `#!/usr/bin/env bash` ?

test/tap-filter
dansimau: cool, I would do this for sure if there was anything else being printed on stderr elsewhere in the script. As it stands, creating a function would mean the reader has to go to another place to trace what is happening.

One other possible way to write it would also be:

```
				{
					echo
					echo -e "\033[1;31m${test_name}\033[0m"
					echo "$buffer"
				} >&2
```

I think I'll just leave it as-is, though.

test/tap-filter
dansimau: Meh. Call me a neckbeard sysadmin, but I prefer `#!/bin/bash`!

test/tap-filter
dansimau: Oh, I see what you mean. no, thanks for spotting!

Motiejus:
looks good to me

PR #31, opened 2016-02-16
fix flappy test where the sut prematurely found out a member was unreachable trough an indirect health check.

CorgiMan:
fix flappy test by making it impossible for the real node to determine that the target node is down

@uber/ringpop 

In ringpop, a node E is declared suspect trough the indirect ping mechanism: When a node A fails to ping E; A randomly picks three members to send a ping-req to, say B, C and D. Upon receiving the ping-reqs the three nodes will ping E and send back the status (ping ok / ping not ok) to A. If all three results are "ping not ok", A will declare E suspect.

In this test we check ringpop's behaviour from the perspective of B (or C or D). We check if B responds with "ping not ok" since E is not responsive. It is important that B doesn't decide that E is a suspect when the ping from B to E fails, which could be a likely programmer mistake. Therefore we check that the all members are alive and there are no suspects in B's membership at the end of the test.

However, sometimes B found out that E is suspect trough it's own ping mechanism, which caused the test to fail. By disabling all pings and ping-reqs, B will never be able to decide if E is unreachable and therefor no longer declares E suspect.

### DIFF
--- a/test/ping-req-tests.js
+++ b/test/ping-req-tests.js
@@ -28,7 +28,19 @@ var getClusterSizes = require('./it-tests').getClusterSizes;
 
 test2('ping-req real-node with a disabled target', getClusterSizes(2), 20000,
     prepareCluster(function(t, tc, n) { return [
-        dsl.disableNode(t, tc, 1),
+        // Our goal in this test is to make a Fake node (F1) do a ping-req
+        // through the Real node (R) to another Fake node (F2), and see if R
+        // correctly performs the ping-req: returns not-ok since F2 is disabled.
+
+        // However, it's possible that R will determine that F2 is dead trough
+        // it's own separate indirect ping mechanism. This will cause R to
+        // declare F2 as suspect, which will interfere with the test.
+
+        // To avoid R declaring F2 a suspect, we disable ping and ping-req in
+        // all Fake nodes, so that R can never perform a successful indirect
+        // health check itself.
+
+        dsl.disableAllNodesPing(t, tc),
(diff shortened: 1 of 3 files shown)


test/ping-req-tests.js
```diff
@@ -28,7 +28,12 @@ var getClusterSizes = require('./it-tests').getClusterSizes;
 
 test2('ping-req real-node with a disabled target', getClusterSizes(2), 20000,
     prepareCluster(function(t, tc, n) { return [
-        dsl.disableNode(t, tc, 1),
+        // By disabling all nodes, it's impossible for the real node to do an
+        // indirect health check itself. This denies the real node to make the
+        // target suspect. This is important because we check that the real
+        // node doesn't change the status of the target when sending a ping-req
+        // ping.
+        dsl.disableAllNodesPing(t, tc),
```
Motiejus: Our goal in this test is to make a Fake node (`F1`) do a `ping-req` through the Real (`R`) node to Fake node (`F2`), and see the Real node correctly does the `ping-req`: returns `ok` if the `F2` is healthy, and `not ok` if `F2` is not healthy.

However, it can happen that Real node will send two more `ping-req` to 2 other Fake nodes (`F3` and `F4`), which will cause `R` to declare `F2` as suspect, which will interfere with the test. We don't want that.

To avoid `R` declaring `F2` a suspect, we can disable `ping` and `ping-req` in `F3` and `F4` (i.e., disable ping in all the Fake nodes).

Motiejus:
LGTM, but please put the PR description to the commit message.

PR #32, opened 2016-02-17
fix flappy "ping-req not ok" test

CorgiMan:
In ringpop, a node E is declared suspect trough the indirect ping mechanism: When a node A fails to ping E; A randomly picks three members to send a ping-req to, say B, C and D. Upon receiving the ping-reqs the three nodes will ping E and send back the status (ping ok / ping not ok) to A. If all three results are "ping not ok", A will declare E suspect.

In this test we check ringpop's behaviour from the perspective of B (or C or D). We check if B responds with "ping not ok" since E is not responsive. It is important that B doesn't decide that E is a suspect when the ping from B to E fails, which could be a likely programmer mistake. Therefore we check that the all members are alive and there are no suspects in B's membership at the end of the test.

However, sometimes B found out that E is suspect trough it's own ping mechanism, which caused the test to fail. By disabling all pings and ping-reqs, B will never be able to decide if E is unreachable and therefor no longer declares E suspect.

rewrite comment

### DIFF
--- a/test/fake-node.js
+++ b/test/fake-node.js
@@ -39,7 +39,7 @@ function FakeNode(options) {
     this.endpoints = {};
     this.enableEndpoints();
 
-    this.pingEnabled = true;
+    this.enabled = true;
 }
 
 FakeNode.prototype.enableEndpoints = function enableEndpoints() {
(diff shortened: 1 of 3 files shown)


Motiejus:
LGTM :+1:

thanodnl:
LGTM

But I am thinking, wouldn't it be better to turn off the gossip of the real node during the test to prevent it from pinging on its own?

CorgiMan:
Agreed @thanodnl. This functionality isn't present in the integration tests yet, and this is a quick and somewhat dirty fix.

PR #33, opened 2016-03-08
Integration tests for reaping faulty nodes.

thanodnl:


test/it-tests.js
```diff
@@ -30,11 +30,105 @@ var clusterSizes = [1, 2, 3, 4, 5, 6, 7, 10, 21, 25, 30];
...
+  memo.push(val);
+  return memo;
+}
+
 function main() {
     program
         .version(require('../package.json').version)
         .option('-s, --sizes <clusterSizes>', 'Cluster sizes to test against. Default: \'' +
              JSON.stringify(clusterSizes) + '\'')
+        .option('--enable-feature <feature>', 'Run tests for experimental features', collect, [])
```
Motiejus: Like discussed in https://github.com/uber/ringpop-go/pull/123, how about running all tests, all the time, but allowing some of them to fail?

test/membership-checksum.js
```diff
@@ -19,29 +19,33 @@
...
-            return 0;
-        }
-    });
+    checksumString = _
+        .chain(members)
+        .filter(function (member) {
+            // remove members that are in the tombstone state
+            return ['tombstone'].indexOf(member.status) < 0;
+        })
+        .sort(function sort(a, b) {
```
Motiejus: Saves a couple of lines. :-)

```javascript
function(x, y){ return a.address > b.address? -1 : x < y ? 1 : 0; },
```

test/reaping-faulty-nodes.js
```diff
@@ -0,0 +1,97 @@
...
+var events = require('./events');
+var test2 = require('./test-util').test2;
+var testStateTransitions = require('./test-util').testStateTransitions;
+var prepareCluster = require('./test-util').prepareCluster;
+var prepareWithStatus = require('./test-util').prepareWithStatus;
+var dsl = require('./ringpop-assert');
+var getClusterSizes = require('./it-tests').getClusterSizes;
+
+test2('join cluster with tombstone in memberlist', getClusterSizes(2), 20000, function init(t,tc, callback) {
+    tc.addExtraMember('192.0.2.100:1234', 'tombstone', 666);
```
Motiejus: Does it have to be a valid ipv4 address? If it's just a string that needs to look like an IP address, can we make something like `256.0.0.0:1` ?

thanodnl: We have settled to use 192.0.2.0/24 for testing ip addresses as described here: https://tools.ietf.org/html/rfc5737

test/it-tests.js
thanodnl: Solved via printing `#: WARNING skipping test suite: ......` and fix the passthrough in `tap-filter`

dansimau: :+1: :smile:

test/reaping-faulty-nodes.js
Motiejus: Nice, I wasn't aware of them.

Motiejus:
LGTM!

test/test-coordinator.js
```diff
@@ -245,6 +247,19 @@ TestCoordinator.prototype.callEndpoint = function callEndpoint(endpoint, body, c
 TestCoordinator.prototype.getMembership = function getMembership() {
     return this.getFakeNodes().map(function(node) {
         return node.toMemberInfo();
+    }).concat(this.extraMembers);
+};
+
+TestCoordinator.prototype.addExtraMember = function(address, status, incarnationNumber) {
```
CorgiMan: AddExtra is a bit redundant, is there a reason to name it like this?

CorgiMan:
Very nice changes, looks good after you have resolved/explained the two comments I have

test/test-coordinator.js
```diff
@@ -245,6 +247,19 @@ TestCoordinator.prototype.callEndpoint = function callEndpoint(endpoint, body, c
 TestCoordinator.prototype.getMembership = function getMembership() {
     return this.getFakeNodes().map(function(node) {
         return node.toMemberInfo();
+    }).concat(this.extraMembers);
+};
+
+TestCoordinator.prototype.extraMember = function(address, status, incarnationNumber) {
```
CorgiMan: Why not addMember? is that already used?
I think that the word extra doesn't convey the adding of the member

thanodnl: I want to distinguish it from the fake nodes. Which are not called members, but I named it like this to make it more clear that we are adding extra information to the membership list instead of starting an other full member of the ring.

CorgiMan: Ahh, ok.
What do you think about addMembershipInformation
else AddExtraMember is fine, just wanted to know the reasoning behind it.

thanodnl: I will use `addMembershipInformation `

CorgiMan:
Looks good after addMembershipInformation change!

PR #35, opened 2016-03-16
Fix redundant if in argument handling

Motiejus:
The legacy function signature is now redundant, we can remove it.


thanodnl:
Good catch, LGTM

thanodnl:
It turned out that there was only 1 call site which passes in the `statusCounts` object so it worked since it was never passed in as numbers. Hence the removal of the block and the arguments ;)

CorgiMan:
looks good

PR #37, opened 2016-03-25
Feature/admin reap

thanodnl:
This add tests for reaping faulty nodes

1. It adds a test for the `/admin/reap` endpoint that can be used to reap all the currently faulty nodes manually.
2. It adds tests that specifically sends `tombstones` as a state and as a flag and checks that ringpop always gossips with the flag variant.

The later test is to make sure that ringpop stays compatible with older versions of `ringpop-go`

### DIFF
--- a/test/reaping-faulty-nodes.js
+++ b/test/reaping-faulty-nodes.js
@@ -121,3 +121,19 @@ test2('tombstone should not be applied when gossiped about but unknown to the SU
         ];
     })
 );
+
+test2('test /admin/reap endpoint', getClusterSizes(2), 20000,
+    prepareWithStatus(1, 'faulty', function(t, tc, n) { return [
+        dsl.assertStats(t, tc, {alive: n, faulty: 1}, {1: {status: 'faulty'}}),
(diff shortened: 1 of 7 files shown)


test/reaping-faulty-nodes.js
```diff
@@ -121,3 +127,107 @@ test2('tombstone should not be applied when gossiped about but unknown to the SU
...
+    })
+);
+
+test2('tombstone should be applied when sent as a flag', getClusterSizes(2), 20000,
+    prepareCluster(function(t, tc, n) {
+        return [
+            dsl.sendPing(t, tc, 0, {
+                sourceIx: 0,
+                subjectIx: 1,
+                status: 'tombstoneFlag'
```
dansimau: Is it possible to somehow just set the `tombstone: true` flag here? It took me a minute to figure out what was going on because 'tombstoneFlag' is not a real status...

dansimau:
LGTM :+1:

severb:
LGTM

PR #39, opened 2016-03-31
Add checksum calculation so that no full sync is performed by SUT

CorgiMan:


test/test-coordinator.js
```diff
@@ -245,9 +245,13 @@ TestCoordinator.prototype.callEndpoint = function callEndpoint(endpoint, body, c
...
 
 TestCoordinator.prototype.getMembership = function getMembership() {
+    var sutHost = this.sutHostPort.split(':')[0]
+    var sutPort = this.sutHostPort.split(':')[1]
+
     return this.getFakeNodes().map(function(node) {
         return node.toMemberInfo();
-    }).concat(this.extraMembers);
+    }).concat(this.extraMembers)
+    .concat({host:sutHost, port:sutPort, status:'alive', incarnationNumber:this.sutIncarnationNumber});
```
Motiejus: I find the SUT membership information to be very useful -- maybe you can put this dictionary to a function in `tc`?

Motiejus:
If I understand correctly, this avoids all unintentional full-syncs by the SUT? If yes, then *awesome job*!  :scream_cat:

CorgiMan:
That's right! 
ringpop-node calculates the checksum a bit differently, but I think we can work around it when we implement the ringpop-node bidirectional full syncs

test/test-coordinator.js
CorgiMan: Good idea

PR #41, opened 2016-04-01
add test/feature for bidirectional full syncs

CorgiMan:


test/bidir-full-sync-tests.js
```diff
@@ -0,0 +1,45 @@
...
+
+            // this causes the SUT to full sync since the chechsums do not match
+            dsl.sendPing(t, tc, 0),
+            dsl.waitForPingResponse(t, tc, 0),
+
+            // the SUT sends a join request to perform a bidirectional full sync
+            dsl.waitForJoins(t, tc, 1),
+
+            // wait for the SUT to receive and process the join response
+            dsl.wait(100),
```
Motiejus: This assumes that in 100ms it will process the join request, which might be false in CI.

Can we instead poll for stats every X (say 10ms), and, after Y times we didn't receive the right answer, fail then? This will speed up the test in the "fast machine" case (when it takes 2ms to process the join request), and will still work in travis.

In that case, we can extend the timeout even more -- to, say, 5 seconds, since on the "good" case it will still take much less.

CorgiMan: Good idea

test/bidir-full-sync-tests.js
```diff
@@ -0,0 +1,45 @@
...
+            dsl.assertStats(t, tc, n+1-3, 0, 3),
+	    ];
+	})
+);
+
+// create partition marks some fake nodes as faulty but doesn't inform the SUT
+// on the changes.
+function createPartition(t, tc) {
```
thanodnl: Can you think about the name of this function? I would not call the behavior you are simulating here a partition.

CorgiMan: That's a fair point. Will rename.

test/fake-node.js
```diff
@@ -91,9 +94,16 @@ FakeNode.prototype._registerEndpoints = function _registerEndpoints() {
         var path = self.endpoints[endpointType].path;
 
         self.channel.register(path, function handleRequest(req, res, arg2, arg3) {
+            self.endpoints[endpointType].handler(req, res, arg2, arg3);
+
+            // After we have handled the request and sent the response, we emit
+            // the request so that the validate function can store and digested
```
Motiejus: s/digested/digest

test/fake-node.js
```diff
@@ -91,9 +94,16 @@ FakeNode.prototype._registerEndpoints = function _registerEndpoints() {
...
+            // it. We want to emit this after, and not before the response has
```
Motiejus: Why do we want to handle the request first, and then call the test handler? Because that's how we write the tests: we poke SUT, SUT reacts, we assert. If you look at the history of this code, you'll see the opposite: we poke SUT, and this happen in parallel:

1. We react (call the test handler).
2. SUT reacts by, e.g., sending messages from the first reaction.

Motiejus:
LGTM

Motiejus:
VLGTM

CorgiMan:
VTX

PR #42, opened 2016-04-06
Partition healing integration tests

Motiejus:
Integration tests against https://github.com/uber/ringpop-go/tree/ws-partition-heal-disco.

Introduces 3 tests:
1. Memberships are incompatible. Verify messages are sent to make them compatible.
2. Memberships are compatible. Verify partitions merge.
3. Two partitions do not know about each other. Verify initial join request is sent.


test/partition-healing-tests.js
```diff
@@ -0,0 +1,377 @@
...
+function decreaseIncOfSutInB(t, tc) {
+    return function decreaseIncOfSutInB(list, cb) {
+        var B = tc.test_state['b1'].membership;
+        _.find(B, {port: tc.test_state['sut'].port}).incarnationNumber = 1000;
+        cb(list);
+    }
+}
+
+
+function splitMemberships(t, tc) {
```
CorgiMan: I think some more comments are appropriate here since I think the name of the function is multi-interpretable. Is the idea that we can change the membership of b1 and the membership of b2 changes at the same time. Can't we just keep them in sync manually to reduce some of the complexity?

test/partition-healing-tests.js
```diff
@@ -0,0 +1,377 @@
...
+function waitForPingResponseFromTarget(t, tc) {
+    return function waitForPingResponseFromTarget(list, cb) {
+        return dsl.waitForPingResponse(t, tc, tc.test_state['target_idx'])(list, cb);
+    }
+}
+
+// Mark b1 and b2 faulty in a1.
+function markBFaultyInA(t, tc) {
+    return function markBFaultyInA(list, cb) {
+        var b1_port = tc.test_state['b1'].port,
```
CorgiMan: At this point I'm a bit confused, when is tc.test_state['b1'] used for tracking membership state and when is tc.fakeNodes[ix_for_b1].membership

test/partition-healing-tests.js
```diff
@@ -0,0 +1,377 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+var TestCoordinator = require('./test-coordinator');
+var _ = require('lodash');
+var safeJSONParse = require('./util').safeParse;
+var dsl = require('./ringpop-assert');
+var events = require('./events');
+var fs = require('fs');
```
CorgiMan: Why is this imported?

test/partition-healing-tests.js
Motiejus: you are right, the previous implementation was terrible. I updated it to have no shared items any more. Thanks!

Motiejus:
Great comments! I addressed your feedback.

test/partition-healing-tests.js
CorgiMan: Can we also change the name to something more clear. "split" sounds a bit too much like "partition" to me

test/partition-healing-tests.js
```diff
@@ -0,0 +1,354 @@
...
+// Assign names to nodes and update membership of B from test coordinator.
+function makeAFaultyInB(t, tc) {
+    return function makeAFaultyInB(list, cb) {
+        tc.fakeNodes[1].status = 'alive';
+        tc.fakeNodes[2].status = 'alive';
+
+        var B = tc.getMembership();
+        // According to B, a1 is faulty
+        _.find(B, {port: tc.fakeNodes[0].port}).status = 'faulty';
+        _.find(B, {port: tc.fakeNodes[0].port}).incarnationNumber = 1337;
```
CorgiMan: Why do we need to change the incarnation number?

test/partition-healing-tests.js
```diff
@@ -0,0 +1,354 @@
...
+                verifySuspects(t, tc, [0], [1, 2])
+            ]
+        }
+        )
+     );
+
+// Assign names to nodes and update membership of B from test coordinator.
+function makeAFaultyInB(t, tc) {
+    return function makeAFaultyInB(list, cb) {
+        tc.fakeNodes[1].status = 'alive';
```
CorgiMan: Why do we need to make them alive, aren't they already in this state?

test/partition-healing-tests.js
```diff
@@ -0,0 +1,354 @@
...
+        _.find(B, {port: tc.fakeNodes[0].port}).incarnationNumber = 1337;
+
+        tc.fakeNodes[1].membership = _.cloneDeep(B);
+        tc.fakeNodes[2].membership = _.cloneDeep(B);
+
+        cb(list);
+    }
+}
+
+function removeHealPartitionDiscoResponse(t, tc) {
```
CorgiMan: Wouldn't the name "waitForHealPartitionDiscoResponse" be more consistent with the rest of the integration tests?

test/partition-healing-tests.js
```diff
@@ -0,0 +1,354 @@
...
+                                change.address + " being alive. Unexpected " +
+                                "merge of partitions!");
+                    }
+                }
+            });
+            // We found all suspects we were looking for in this message
+            if (_.isEqual(addresses.sort(), suspects.sort())) {
+                t.ok(true, "Found " + nodes + " suspects (" + suspects + ") we were looking for");
+                _.remove(list, msg);
+                return cb(list);
```
CorgiMan: the early return confuses me a bit, since we are also checking for an Unexpected merge of partitions, shouldn't we check all the ping requests?

CorgiMan: It also feels a bit arbitrary to remove the ping that contains the last suspect declaration. I think not removing any, or removing every ping would be more appropriate

test/partition-healing-tests.js
Motiejus: No, they became faulty in lines 45-46:

```
   45                 dsl.changeStatus(t, tc, 1, 1, 'faulty', 0),
   46                 dsl.changeStatus(t, tc, 2, 2, 'faulty', 0),
```

test/partition-healing-tests.js
Motiejus: This is a leftover from pre-pre-previous version. Good question, removed.

test/partition-healing-tests.js
Motiejus: :+1:

test/partition-healing-tests.js
CorgiMan: I think we are looking for a single ping that contains the suspects and alive declarations from the other partition. Now we are accumulating the suspects.

test/partition-healing-tests.js
```diff
@@ -0,0 +1,354 @@
...
+            cb(null);
+        } else {
+            list.splice(list.indexOf(d), 1);
+            cb(list);
+        }
+    }
+}
+
+// Verify 'nodes' are treated as suspects by sut. Also, verify sut is not
+// sending 'nodes' information that 'nodes_in_other_partition' are alive.
```
CorgiMan: Can we add a comment along the lines of: verifySuspect waits for a ping from the sut that contains the following changes: nodes from nodes are declared suspect, and no nodes from nodes_in_other_partition are declared alive.

CorgiMan:
Great work on the integration tests! This is by far the most involved feature that we are testing.

test/partition-healing-tests.js
CorgiMan: Ah right! thanks

test/partition-healing-tests.js
Motiejus: This function had multiple problems, you identified them really well. I simplified the function and fixed the problems. Good catch.

test/partition-healing-tests.js
Motiejus: :+1: added.

test/partition-healing-tests.js
```diff
@@ -307,8 +309,8 @@ function decreaseIncOfSutInB(t, tc) {
 }
 
 
-function splitMemberships(t, tc) {
-    return function splitMemberships(list, cb) {
+function separateMemberships(t, tc) {
```
CorgiMan: Still not too happy with this. What do you think of localizeMemberships, since that is the word you use in the comment to describe the call .

CorgiMan:
👍  Looks good!

test/partition-healing-tests.js
```diff
@@ -0,0 +1,357 @@
...
+// ● a1, sut: partition A.
+// ● b1, b2: partition B.
+//
+// 1. Create two partitions A and B by marking b1 and b2 failed in sut.
+//    a1 and sut are failed according to B.
+// 2. Trigger partition healing.
+// 3. Verify sut tries to reincarnate both sides:
+// 3.1. sut declares b1 and b2 as suspects to B.
+// 3.2. sut declares a1 as suspect to A.
+test2('reincarnating partitions A and B', [3], 20000,
```
thanodnl: should we run this on all clusters bigger than 3? currently you pass in `[3]`, most tests use [`getClusterSizes`](https://github.com/uber/ringpop-common/blob/master/test/it-tests.js#L186) to select the sizes that are passed in via the cli that are bigger then their minimum size.

test/partition-healing-tests.js
```diff
@@ -0,0 +1,357 @@
...
+
+        var pingRequests = _.filter(list, {
+            type: events.Types.Ping,
+            direction: 'request'
+        });
+
+        var found_all_suspects = false;
+        _.forEach(pingRequests, function(pingRequest) {
+            var suspects = [];
+            _.forEach(safeJSONParse(pingRequest.arg3).changes, function(change) {
```
thanodnl: instead of safeparsing you can use `pingRequest.body`. It is already safe parsed in the `events.js`

thanodnl:
Overal looks great, but I wonder if it would make sense to also add a test for 2 partitions that do not know about each other. Be it caused by wrongfully bootstrapping, or a network partition that is longer than the `faulty period` + `tombstone period`.

test/partition-healing-tests.js
Motiejus: Good one. I thought about this initially, but it was really hard to maintain the state even on 3 nodes. Now as I split them out better, hopefully this gets easier.

I'll work on this.

test/partition-healing-tests.js
Motiejus: Thanks, good to know.

Motiejus:
@thanodnl thank you for your feedback. I`ll try to address all the points.

test/partition-healing-tests.js
Motiejus: I implemented `healerUnknownNodes` with arbitrary number of fake nodes (with minimum being 2). That was a pretty simple test, and it worked perfectly. While I agree with your comment that parameterizing the cluster size would match the style of the rest of the tests, I don't think the value is worthwhile for added complexity.

The biggest problem in the other two tests is splitting and maintaining the membership information of fake nodes. It is not trivial with 3 nodes already, so it's going to be even more difficult with `n`.

Do you think there is another argument besides being consistent with the rest of the code base? If not, I would rather not add even more complexity to the two tests which have this non-parameterized.

thanodnl: If implementing this test with an arbitrary cluster size is too hard let make sure that this test then only runs when we are testing the cluster size of 3. For the integration tests we spawn a new integration test for every size we are testing.

Motiejus: You mean the test should be skipped iff `3` is not in the list of cluster sizes? I.e. change from "run this test always" to "run this test iff 3 is an option"?

I am happy to do the change, but can you explain why it's better than it is now? I think there's a risk that if we ever change the default to not include 3, these test would be unintentionally skipped.

CorgiMan:
Is this ready to land @thanodnl @Motiejus ?

test/partition-healing-tests.js
```diff
@@ -395,7 +486,7 @@ function addMoreFakeNodes(t, tc, n) {
 }
 
 function validateHealRequest(t, tc, eventBody) {
-    t.equal(1, eventBody['targets'].length,
-            "/join was sent to one node as part of heal");
+    t.equal(eventBody['targets'].length, 1,
+            "expected 1 join target as part of the heal");
```
CorgiMan: Is this what we discussed (heal should stop upon one fail?)

The new proposal is to stop after 10 failures to offer some resilience against nodes that are truely faulty.

test/partition-healing-tests.js
```diff
@@ -0,0 +1,492 @@
...
+        // Start n-1 new fake nodes.
+        async.each(tc.fakeNodes.slice(1), function startNode(node, nodeStarted) {
+            node.start(nodeStarted);
+        }, function() {
+            cb(list);
+        });
+    }
+}
+
+function validateHealRequest(t, tc, eventBody) {
```
CorgiMan: I think this should be called, validateHealResponse

PR #43, opened 2016-04-08
Making a local partition

Motiejus:
1. Add a script how to make a local partition.
2. Update documentation to match the script.

### DIFF
--- a/tools/make_partition
+++ b/tools/make_partition
@@ -13,6 +13,16 @@ Clear the firewall:
 Check current firewall rules:
    pfctl -sr               # Darwin
    iptables -nvL           # Linux
+
+=== Limitations ===
+Undefined behavior when something else is listening on the same port, but
+different IP address, e.g.
+
+    pop  1  r  3u  IPv4 84  0t0  TCP 127.0.0.1:3000 (LISTEN)
+    met  2  m  3u  IPv4 84  0t0  TCP 10.0.0.10:3000 (LISTEN)
+
+If the above holds true, it can break firewall in spectacular ways. So not
+intended for unattended use.
 """
 
 import sys
--- a/tools/make_partition
+++ b/tools/make_partition
@@ -93,7 +93,7 @@ def _block_tuples(lsof, port_to_owner):
 
 def _block_preamble_darwin(ip, ports):
     """Block new connections to listening ports on Darwin."""
-    tpl = "block return-rst proto tcp from %s port %s flags S/S"
+    tpl = "block return in proto tcp from %s port %s flags S/S"
     return [tpl % (ip, port) for port in ports]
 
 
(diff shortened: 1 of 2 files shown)


tools/make_partition
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env python
+"""
+Usage:
+    sudo lsof -Pnni | ./make-partition 3000,3001 3002,3003 | sudo FIREWALL
```
CorgiMan: typo: make_partition

tools/make_partition
```diff
@@ -0,0 +1,204 @@
...
+
+FIREWALL :=
+    pfctl -emf -      # Darwin
```
CorgiMan: What is the trailing dash for?

Motiejus: `-f` stands for filename, and `-` is a standard for standard input.

tools/make_partition
```diff
@@ -0,0 +1,214 @@
...
+    """iptables-save requires "COMMIT" at the end."""
+    return ["COMMIT"]
+
+
+def main():
+    parser = argparse.ArgumentParser(
+            formatter_class=argparse.RawDescriptionHelpFormatter,
+            description='Generate firewall rules to create a partition.',
+            epilog=__doc__)
+    parser.add_argument('A', help='Ports of A, e.g. 3000,3001-3002')
```
CorgiMan: is the dash supposed to specify a range? I couldn't get it to work..

Motiejus: Yes, a range. Worked for me:

```sh
sudo lsof -Pnni | ./tools/make_partition 3000-3003,3008 3004-3007 | sudo pfctl -emf -
```

And output from ringpop-admin partitions:

```
 11:28:31.138   Checksum     # Nodes   # Alive   # Suspect   # Faulty   Sample Host
                91644736     5         5         0           4          127.0.0.1:3000
                1106031481   4         4         0           5          127.0.0.1:3004```


Can you clarify what didn't work?

dansimau:
Awesome work! I managed to get this working.

At first it wasn't obvious to me that I had to have the cluster running before the script would run. Also `ringpop-admin top` stops working while the rules are in effect. After I verified the partition was created, I cleared them to get ringpop-admin working.

The docs are a little light on the steps required to actually make a partition and probably having step-by-step commands to copy/paste would alleviate any confusion (imagine someone is totally new to ringpop and they want to create a partition).

Also, the `docs/partitions.md` might be a little out-of-date after the addition of this script:

> _With the current implementation of tick-cluster, it is hard, or even impossible
>  to form a partition with firewall rules alone._
https://github.com/uber/ringpop-common/pull/43/files#diff-95f6b25b1a0c56fb7099d5ae22ef3d4dL36

Aside from that, LGTM.

CorgiMan:
Looks good! Love it.

Motiejus:
@dansimau thank you for your great feedback, indeed the documentation was lacking. I updated it, hopefully you'll like it better.

`ringpop-admin top` and `ringpop-admin partitions` work in partitioned mode too; it's just important to open them before forming the partition (so connections are established). I also put this information to the docs.

dansimau:
ship it

PR #45, opened 2016-04-14
Detect checksum method used by SUT

mennopruijssers:
There is an incompatibility in the way the checksum is calculated between ringpop-node and ringpop-go. Also, the ringpop-node has two ways of hashing the checksum string.

After the SUT is launched, it's `admin/stats`-endpoint is hit to get it's membership list and checksum. Then the correct hashing method is determined by calculating and comparing a checksum based on it's membership list with all three methods. From that point onwards, the correct checksum method is used for calculating checksums in fake-node.

Tested against:
✅ ringpop-node#master
✅ ringpop-node#uber/menno/bi-dir-sync (with `--enable-feature bidirectional-full-syncs`)
✅ ringpop-go#master (with `--enable-feature bidirectional-full-syncs`)
✅ ringpop-go#dev (with `--enable-feature bidirectional-full-syncs`)

 and current dev of ringpop-go without issues.

### DIFF
--- a/test/membership-checksum.js
+++ b/test/membership-checksum.js
@@ -22,22 +22,52 @@
 var _ = require('lodash');
 var farmhash = require('farmhash');
 
-// entries must have address (hostport), status (e.g. "alive"), and incarnation numbers
-module.exports.checksum = function checksum(members) {
-    checksumString = _
+function generateChecksumString(members) {
+    return _
         .chain(members)
         .filter(function (member) {
             // remove members that are in the tombstone state
             return ['tombstone'].indexOf(member.status) < 0;
         })
-        .map(function (member){
-            return member.host + ':' + member.port +
-                member.status +
-                member.incarnationNumber+';';
+        .map(function (member) {
+            var address = member.address || (member.host + ':' + member.port);
+            return address +
+                    member.status +
+                    member.incarnationNumber
         })
         .value()
         .sort()
-        .join('');
-
+        .join(';');
+}
+// entries must have address (hostport), status (e.g. "alive"), and incarnation numbers
+function checksumGo(members) {
+    //add extra semi-colon to be compatible with the go implementation
+    var checksumString = generateChecksumString(members) + ';';
     return farmhash.fingerprint32(checksumString);
+}
+
+function checksumNode(members) {
+    var checksumString = generateChecksumString(members);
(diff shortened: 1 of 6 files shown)


dansimau:
LGTM

test/membership-checksum.js
```diff
@@ -22,22 +22,52 @@
...
+    } else if (checksumNodeCrossPlatform(members) === checksum) {
+        return checksumNodeCrossPlatform;
+    } else {
+        throw new Error('checksum method undetectable!');
+    }
+}
+
+module.exports = {
+    detect: detect
+};
```
thanodnl: nit: new line at end of file

thanodnl:
LGTM

test/test-coordinator.js
```diff
@@ -316,4 +316,9 @@ TestCoordinator.prototype.getSUTHostPort = function getSUTHostPort() {
     return this.sutHostPort;
 };
 
+TestCoordinator.prototype.checksum = function checksum() {
+    //default to no-checksum. This method is changed based on runtime data of the SUT
+    return 0;
```
Motiejus: `0` is a valid checksum value which can get exposed to the test. To make it clear that this is not an actual checksum, but a mock implementation, should we change it to either `-1`, or, if we want to make sure this function is always overwritten, raise an exception?

test/test-coordinator.js
```diff
@@ -316,4 +316,9 @@ TestCoordinator.prototype.getSUTHostPort = function getSUTHostPort() {
     return this.sutHostPort;
 };
 
+TestCoordinator.prototype.checksum = function checksum() {
+    //default to no-checksum. This method is changed based on runtime data of the SUT
```
Motiejus: Can you add the description of the pull request to the line above this function signature? In that case, we'll know very well what this method is used for.

test/test-coordinator.js
mennopruijssers: Raise an exception isn't possible since the SUT will already start joining the fakeNodes before we had a chance to detect the checksum method. I'll change it to `null` to be more clear.

mennopruijssers:
Rebased and addressed the comments. Also ran the `reaping-faulty-nodes` test against `ringpop-go#feature/reaping-faulty-nodes`.

PR #47, opened 2016-04-20
Add throttle test for bidirectional full syncs

mennopruijssers:
As a safety check, both the ringpop-go and ringpop-node limit the number (defaults to 5) of concurrent bidirectional full syncs. 
A bidirectional full sync is triggered by a ping to a node without outstanding membership changes where the checksum in the ping doesn't match the node's checksum. This results in the node sending it's full membership. At the same time, the node will send a join to the source of the ping request to get it's full membership list.
This test pauses responding to the join request to make sure there are multiple joins happening at the same time. If the throttling is working, there should never be more than 5 joins in the queue.

### DIFF
--- a/test/bidir-full-sync-tests.js
+++ b/test/bidir-full-sync-tests.js
@@ -67,97 +67,78 @@ test2('bidirectional full sync throttling test', getClusterSizes(3), 20000,
             dsl.assertStats(t, tc, n+1, 0, 0),
             dsl.waitForEmptyPing(t, tc),
 
-            //pause handling joins to stall reverse full syncs
-            overwriteJoinHandlerWithPause(tc, 0),
-
-            // trigger the reverse full syncs
-            triggerReverseFullSyncs(tc, 0, numberOfFullSyncsToTrigger),
-
-            // verify the number of )cached) join requests
-            verifyCachedJoins(t, tc, 0, numberOfFullSyncsToTrigger, expectedJoins),
-
-            // handle the joins and restore the original join-handler.
-            handleCachedJoinsAndRestoreJoinHandler(tc, 0),
+            assertReverseFullSyncThrottling(t, tc, 0, numberOfFullSyncsToTrigger, expectedJoins),
 
             // wait for the joins to be handled.
             dsl.waitForJoins(t, tc, expectedJoins),
-
+            dsl.expectOnlyPings(t, tc),
             dsl.assertStats(t, tc, n+1, 0, 0)
         ];
     })
 );
 
-function overwriteJoinHandlerWithPause(tc, idx) {
-    return function overwriteJoinHandlerWithPause(list, cb) {
-        var fakeNode = tc.fakeNodes[idx];
 
-        fakeNode.cachedJoins = [];
-        fakeNode.originalJoinHandler = fakeNode.endpoints['Join'].handler;
-        fakeNode.endpoints['Join'].handler = function() {
-            fakeNode.cachedJoins.push(arguments)
-        };
+function assertReverseFullSyncThrottling(t, tc, nodeIx, numberOfFullSyncsToTrigger, expectedJoins){
 
-        cb(list);
-    }
-}
+    var cachedJoins;
+    var originalJoinHandler;
 
-function verifyCachedJoins(t, tc, idx, fullSyncs, expectedJoins) {
-    return function verifyCachedJoins(list, cb) {
-        var fakeNode = tc.fakeNodes[idx];
-        var cachedJoins = fakeNode.cachedJoins;
-        if (cachedJoins.length < expectedJoins) {
-            cb(null);
-            return;
-        }
-        t.equal(cachedJoins.length, expectedJoins);
+    return [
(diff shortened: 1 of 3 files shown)


test/bidir-full-sync-tests.js
```diff
@@ -47,6 +48,107 @@ test2('bidirectional full sync test', getClusterSizes(5), 20000,
...
+            assertReverseFullSyncThrottling(t, tc, 0, n*3, 5),
+
+            dsl.assertStats(t, tc, n+1, 0, 0),
+        ];
+    })
+);
+
+// assert if the number of joins is throttled.
+function assertReverseFullSyncThrottling(t, tc, idx, count, expectedJoins) {
+    var functions = [];
```
CorgiMan: I prefer if all the functions are just declared within the [] brackets, since I think it improves readability and is more in line with the rest of the integration tests.

test/bidir-full-sync-tests.js
```diff
@@ -47,6 +48,107 @@ test2('bidirectional full sync test', getClusterSizes(5), 20000,
...
+            }, function done() {
+                cb(list);
+            });
+        };
+        f.callerName = 'triggerReverseFullSyncs';
+
+        return f;
+    }
+
+    function pauseHandlingJoins(list, cb) {
```
CorgiMan: I like the name overwriteJoinHandlerWithPause or something similar better. Or some indication that we overwrite the join handler here

test/bidir-full-sync-tests.js
```diff
@@ -47,6 +48,107 @@ test2('bidirectional full sync test', getClusterSizes(5), 20000,
...
+
+    //pause handling joins to stall reverse full syncs
+    functions.push(pauseHandlingJoins);
+    functions.push(dsl.waitForEmptyPing(t, tc));
+
+    // trigger the reverse full syncs
+    functions.push(triggerReverseFullSyncs(tc, idx, count));
+
+    // verify the number of join requests
+    functions.push(verify);
```
CorgiMan: there is dsl.waitForJoins(x). I think that does exactly what you need

CorgiMan:
Good job man, color me impressed.

test/bidir-full-sync-tests.js
mennopruijssers: I'm using `dsl.waitForJoins` later. In the verify method I just verify the number of cached join calls. They won't show up in waitForJoins because I had to intercept them.

test/bidir-full-sync-tests.js
mennopruijssers: In a previous version this wasn't possible because of a for-loop but after I restructured it a bit it's possible. I'll change it.

CorgiMan: Sorry, I meant just replace 
```
var functions = [];
functions.push( X );
functions.push( Y );
```
by
```
var functions = [ 
  X,
  Y,
]
```
doesn't that work?

test/bidir-full-sync-tests.js
CorgiMan: Hmm, you're right. Every handler that is registered before this was registered with _registerHandlers. That function adds some magic sauce on the handlers that pushes the every incoming request on the events list.
Can you refactor that so that it uses _registerHandler (singular) and use that function here, in stead of overwriting the handler manually. Then you can use, waitForJoins, and I think if you do that, you don't need to restore the handlers again either.

dansimau:
LGTM, but as discussed, it would be nice to see this running on Travis. Let's ship it and see what happens...

test/bidir-full-sync-tests.js
mennopruijssers: It's not possible because the we still want to "hold" the join-requests and reply to them later. Good point about the secret sauce; I changed the overwrite and restore a little bit to make sure we don't loose it.

test/bidir-full-sync-tests.js
mennopruijssers: Should have worked but do you prefer that over the current way of listing them all in the main test-method?

CorgiMan:
Can you use test_state in testCoordinator to store the cachedJoins please. Sorry for not mentioning that earlier.

CorgiMan:
Sorry I have a few more things...

test/ringpop-assert.js
```diff
@@ -224,13 +224,13 @@ function changeStatus(t, tc, sourceIx, subjectIx, status, subjectIncNoDelta) {
     return f;
 }
 
-function sendPing(t, tc, nodeIx, piggybackOpts) {
+function sendPing(t, tc, nodeIx, piggybackOpts, bodyOverrides) {
```
Motiejus: As It is getting hard to understand what the function is doing from the docstring + signature, please document `piggybackOpts` and `bodyOverrides`.

Motiejus:
I like your approach on putting all the internal test data to a closure. Nice!

Apart from my above non-blocking comment about function documentation, LGTM.

CorgiMan:
Looks good!

PR #48, opened 2016-04-20
use more idiomatic functions from ringpop-assert

CorgiMan:
This shows the use of functions that are more idiomatic in the integration tests. I'm not sure if it tests everything that you wanted Menno, but I'm content with testing this much.


mennopruijssers:
I guess this one can be closed since I modified #47 and merged that one back in master.

PR #50, opened 2016-05-03
Increase piggyback test timeout

severb:
This fixes a timeout in the integration tests after the addition of reaping faulty nodes. Because of the increased piggyback counter when the custersize >= 10 to 30, the dissemination takes 6s, but we transition from faulty to tombstone after 5s. This adds another round of gossiping.


thanodnl:
LGTM

CorgiMan:
Looks good

severb:
❤️

PR #51, opened 2016-05-03
Fix when running integration test for more than one cluster size

CorgiMan:
Some global state was changed which broke running the integration tests for multiple cluster sizes (statusCounts was mutated by the testStateTransition function). This PR creates a copy of statusCounts to avoid the mutation of statusCounts.

### DIFF
--- a/test/test-util.js
+++ b/test/test-util.js
@@ -97,8 +97,8 @@ function testStateTransitions(ns, initial, newState, finalState, incNoDelta, sta
             expectedMembers[ix] = {status: finalState};
 
             // alive is a delta
-            counts = _.extend({}, statusCounts)
-            counts.alive = (counts.alive || 0) + n
+            var counts = _.extend({}, statusCounts);
+            counts.alive = (counts.alive || 0) + n;
             return [
                 dsl.changeStatus(t, tc, 0, 1, newState, incNoDelta),
                 dsl.waitForPingResponse(t, tc, 0),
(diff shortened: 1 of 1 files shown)


test/test-util.js
```diff
@@ -95,11 +97,12 @@ function testStateTransitions(ns, initial, newState, finalState, incNoDelta, sta
             expectedMembers[ix] = {status: finalState};
 
             // alive is a delta
-            statusCounts.alive = (statusCounts.alive || 0) + n
+            counts = _.extend({}, statusCounts)
```
Motiejus: Why is `counts` global?

Honestly, I don't understand this change. What broke and why?

mennopruijssers: Missing `var`?

test/test-util.js
```diff
@@ -95,11 +97,12 @@ function testStateTransitions(ns, initial, newState, finalState, incNoDelta, sta
...
+            counts.alive = (counts.alive || 0) + n
```
mennopruijssers: missing semicolon.

mennopruijssers:
If I understand it correctly, the `prepareWithStatus` callback function is called for each cluster size and it's reading and writing to/from the same `statusCounts`-object? If that is correct: LGTM (besides the two style issues) but please add it as a description of the PR/commit message

CorgiMan:
That's correct Menno

mennopruijssers:
LGTM but again: please make sure the title and description of the PR and commit describe it a little bit better.

thanodnl:
LGTM

PR #52, opened 2016-05-03
Speed up integration tests; 10:18 to 3:12 minutes

CorgiMan:
Before waiting for the disseminator of the real node to drain we send it 30 pings (this will certainly drain it immediately). That means we no longer have to wait 6 seconds for the disseminator to drain every waitForEmptyPing is called.

It is certainly not the cleanest implementation but I really do not want to spent a lot of time on this.

The speedup in from 10:18 to 3:12 minutes. Thats a speedup of **3.2x**.

### DIFF
--- a/test/piggyback-tests.js
+++ b/test/piggyback-tests.js
@@ -28,7 +28,7 @@ var getClusterSizes = require('./it-tests').getClusterSizes;
 test2('ringpop sends piggyback info in ping request', getClusterSizes(3), 20000, prepareCluster(function(t, tc, n) {
     return [
         // TODO clear the dissemination information from the SUT by flooding it with pings instead of waiting for it
-        dsl.waitForEmptyPing(t, tc), // problem is that if decay is not working you might never get to this point
+        dsl.drainDisseminator(t, tc), // problem is that if decay is not working you might never get to this point
 
         dsl.changeStatus(t, tc, 0, 1, 'suspect'),
         dsl.waitForPingResponse(t, tc, 0),
@@ -51,7 +51,7 @@ test2('ringpop sends piggyback info in ping request', getClusterSizes(3), 20000,
 test2('ringpop updates its dissimination list on pingreq', getClusterSizes(3), 20000, prepareCluster(function(t, tc, n) {
     return [
         // TODO clear the dissemination information from the SUT by flooding it with pings instead of waiting for it
-        dsl.waitForEmptyPing(t, tc), // problem is that if decay is not working you might never get to this point
+        dsl.drainDisseminator(t, tc), // problem is that if decay is not working you might never get to this point
 
         // send information to be piggy backed via pingreq
         dsl.sendPingReq(t, tc, 0, 2, {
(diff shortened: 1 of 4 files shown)


test/piggyback-tests.js
```diff
@@ -112,8 +112,8 @@ test2('ringpop piggybacking decays', getClusterSizes(3), 40000, prepareCluster(f
         dsl.waitForEmptyPing(t, tc), // problem is that if decay is not working you might never get to this point
 
         // send information to be piggy backed
-        dsl.changeStatus(t, tc, 0, 1, 'suspect'),
-        dsl.waitForPingResponse(t, tc, 0),
+        dsl.changeStatus(t, tc, 1, 2, 'suspect'),
```
CorgiMan: Need to send ping from different node because the SUT filters out changes that originated from the node. The suspect change now originates from node 1, and node 0 sends the pings to drain the SUT.

test/ringpop-assert.js
```diff
@@ -224,16 +233,11 @@ function changeStatus(t, tc, sourceIx, subjectIx, status, subjectIncNoDelta) {
...
- * @param {Test} t The current running test
- * @param {TestCoordinator} tc The test coordinator.
- * @param {number} nodeIx The index of the fakeNode that should send the ping.
- * @param {object} piggybackOpts The changes to piggy back on the ping.
- * @param {object} bodyOverrides Overwrite specific ping body parameters (checksum, source, sourceIncarnationNumber, changes)
- * @returns {Function} Returns the function that'll be invoked when running the integration test.
- */
+
+// Send a ping to the SUT from a fake node, where:
+// - nodeIx is the index of the fakeNode that should send the ping.
```
jwolski: The deleting of the comment format is typically not a cool thing to do (unless it was decided that we're not going that route).

CorgiMan: Oh sorry, wasn't aware of that etiquette. I thought it would be ok to enforce the consistency of the comments in this way.

test/ringpop-assert.js
```diff
@@ -92,6 +92,15 @@ function waitForPing(t, tc) {
     };
 }
 
+function fasterWaitForEmptyPing(t, tc) {
```
Motiejus: You should document why you are doing this, as the name of the function is not enough. Specifically:

1. Why are you sending pings to the SUT? I know that this will cause the disseminator to drain, but it's not obvious. You should comment that.
2. Why 30? I guess this has something to do with the p-value, but I am not sure. You should also give a reason for that number here.

test/ringpop-assert.js
```diff
@@ -910,8 +914,9 @@ module.exports = {
     waitForJoins: waitForJoins,
     waitForPingReqs: waitForPingReqs,
     waitForPing: waitForPing,
-    waitForEmptyPing: waitForEmptyPing,
-    // drainSUTDissemination: drainSUTDissemination,
+    // waitForEmptyPing: waitForEmptyPing,
+    waitForEmptyPing: fasterWaitForEmptyPing,
```
Motiejus: If we really don't need `waitForEmptyPing`, we should properly rename the exported function. Since no outside projects depend on it, I think we can just do it.

Motiejus:
I like the idea of speeding the integration tests a lot. :-) A few minor things commented, apart from that, lookin' good.

thanodnl:
I like the idea behind the change. I like even more the speedup you achieved with this change. Nice work.

test/ringpop-assert.js
```diff
@@ -910,8 +924,11 @@ module.exports = {
     waitForJoins: waitForJoins,
     waitForPingReqs: waitForPingReqs,
     waitForPing: waitForPing,
-    waitForEmptyPing: waitForEmptyPing,
-    // drainSUTDissemination: drainSUTDissemination,
+
+    // waitForEmptyPing is slow. If possible, use drainDisseminator instead.
+    // waitForEmptyPing: waitForEmptyPing,
```
thanodnl: Lets remove this dead code before merging. The comment above is an actual good comment for the line below so that should stay.

CorgiMan: Yeah that's why I left it in actually, can remove though

thanodnl:
LGTM

Motiejus:
LGTM

PR #53, opened 2016-05-06
Update /admin/reap to return reaped nodes count

severb:
See https://github.com/uber/ringpop-node/pull/257#discussion_r62028527 for context.

### DIFF
--- a/test/reaping-faulty-nodes.js
+++ b/test/reaping-faulty-nodes.js
@@ -174,7 +174,11 @@ test2('test /admin/reap endpoint', getClusterSizes(2), 20000,
             type: events.Types.AdminReap,
             direction: 'response'
         }, 'Wait for AdminReap response', function (response) {
-            return response.body.status === 'ok';
+            var correctCount = true;
+            if (response.body.reaped !== undefined) {
+                correctCount = response.body.reaped === 1;
+            }
+            return response.body.status === 'ok' && correctCount;
         }),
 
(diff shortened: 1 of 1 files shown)


dansimau:
Can you elaborate more in the description? I don't understand what this is doing. It doesn't seem like it is doing what the title says.

severb:
It modifies the /admin/reap test to expect a count instead of `ok`.

dansimau:
Great. Would you mind adding it to the description with a bit of context?

I'm on the war path to get people to write good descriptions because it helps with releases. It only takes 20 seconds to do this and really helps everyone with reviewing or keeping an eye on the projects... 😄

severb:
Totally agree, I added a link for more context.

dansimau:
Doesn't this break the integration test for ringpop-go?
https://github.com/uber/ringpop-go/blob/dev/swim/handlers.go#L162

severb:
It does if you enable testing for the reaping-faulty-nodes feature. We'll do a similar change in ringpop-go before this feature is enabled by default.

dansimau:
Am I missing something? I thought it was already enabled by default in ringpop-go...
https://github.com/uber/ringpop-go/blob/master/test/run-integration-tests#L109

severb:
Oh, you're right! Good catch, I'll need to update this PR.

thanodnl:
LGTM

dansimau:
LGTM

severb:
Thanks!!

PR #54, opened 2016-05-06
Mark bidirectional-full-syncs as mandatory

mennopruijssers:
Because uber/ringpop-node#251 added bidirectional full syncs in ringpop-node, this test should be mandatory moving forward.


Motiejus:
LGTM

dansimau:
LGTM

CorgiMan:
Nice, can't wait till everything is mandatory :)

PR #55, opened 2016-05-06
Make partition-healing tests less flaky

mennopruijssers:


dansimau:
LGTM.

Not blockers, however:

Would recommend a comment to explain why these conditions have been added, e.g. `"// responses can sometimes come back out-of-order on slow machines"`, but not a blocker.

Also, if you are going to make any changes, there is a typo in the commit message: "partition".

Thanks for this, everyone loves less flappy tests 👍

PR #56, opened 2016-05-09
Verify if source-field of changes is not triggering a reverse full-sync

mennopruijssers:
During partition healing the changes send to target of the heal-operation are constructed in such a way that unmerge-able nodes in the memberships are marked as suspect so they will reincarnate. If the source of those changes is equal to the host-port of the node triggering the heal-operation, this could trigger a reverse full-sync. The reverse full-sync subsequently could result in a state where the two partitions start communicating before all the unmerge-able nodes are reincarnated. This would result in a lot of unnecessary gossiping.

### DIFF
--- a/test/partition-healing-tests.js
+++ b/test/partition-healing-tests.js
@@ -370,6 +370,11 @@ test2('dont\'t merge partitions when B is not fully reincarnated', [3], 20000, p
                 return false;
             }
 
+            if (_.filter(ping.body.changes, {source: tc.sutHostPort}).length !== 0) {
+                t.fail("expected the source address to be different to prevent a bi-directional full sync");
+                return;
+            }
+
             return true;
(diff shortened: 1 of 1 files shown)


Motiejus:
Please say why it is a problem in the commit description (IIRC, we have a policy to have it non-empty); that should update the summary of the PR too. Other than that, LGTM.

mennopruijssers:
@Motiejus I updated the title/description of this PR. when merging I'll make sure the commit message would be the same. Is it clear now?

thanodnl:
LGTM, did you run the integration tests against `ringpop-go` with this change?

mennopruijssers:
@thanodnl I did (and they pass ;))

Motiejus:
LGTM

PR #57, opened 2016-05-10
Add heal-stats to running_ringpop.md#monitoring

mennopruijssers:
Add the emitted stats during partition healing to the list with emitted stats.


Motiejus:
LGTM

PR #58, opened 2016-05-12
Add partition healing details

Motiejus:
Implementation is settled, time to make documentation available.

### DIFF
--- a/docs/make-partition.md
+++ b/docs/make-partition.md
@@ -144,13 +144,13 @@ If you are happy with the output, apply the rules:
 OS X:
 
 ```shell
-$ sudo lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003 | sudo pfctl -emf -
+$ sudo -v && sudo lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003 | sudo pfctl -emf -
 ```
 
 Linux:
 
 ```
-$ sudo lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003 | sudo iptables-restore
+$ sudo -v && sudo lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003 | sudo iptables-restore
 ```
 
 In a few seconds, you should see output from `tick-cluster` that some of the
--- a/docs/make-partition.md
+++ b/docs/make-partition.md
@@ -183,7 +183,8 @@ That's it, we have a partition! To break it, we need to wipe the firewall rules:
 * OS X: `pfctl -f /etc/pf.conf`.
 * Linux: `iptables -F`.
 
-... and wait for partition healing to kick in.
+... and wait for partition healing to kick in, or execute the `ringpop-admin
+heal` [command][1].
 
 ## Final remarks
 
@@ -192,3 +193,5 @@ That's it, we have a partition! To break it, we need to wipe the firewall rules:
 * `tools/make_partition` is not intended to be used in an automated way. See
   `--help` to learn about the limitations.
 
+
+[1]: https://github.com/uber/ringpop-admin
--- a/docs/make-partition.md
+++ b/docs/make-partition.md
@@ -1,0 +1,194 @@
+# Forming a partition
+
+With the current implementation of tick-cluster, it is non-trivial to form a
+partition. To understand why, we need to understand how connections are
+established.
+
+## Port Allocation
+
+A ringpop instance opens a local tchannel socket (=listening tcp socket) to
+accept incoming connections from other ringpops. By default, on a 2-node
+tick-cluster, this is `127.0.0.1:3000`. Let's call it instance `a`. For
+instance `a` to establish a connection to instance `b` (`127.0.0.1:3001`),
+instance `a` will open an ephemeral port, e.g. `43323`, to connect to instance
+`b`. This connection, from `127.0.0.1:43323` (`a`) to `127.0.0.1:3001` (`b`) is
+used for messages initiated by node `a`. The other connection (example below),
+from `127.0.0.1:36113` (`b`) to `127.0.0.1:3000` (`a`), is used for messages
+initiated by `b`. Here is a snapshot of `lsof` from a two-node cluster:
+
+```
+root:/# lsof -Pnni
+COMMAND PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
+node     64 root   10u  IPv4 217924      0t0  TCP 127.0.0.1:3000 (LISTEN)
+node     64 root   11u  IPv4 217925      0t0  TCP 127.0.0.1:43323->127.0.0.1:3001 (ESTABLISHED)
+node     64 root   12u  IPv4 217926      0t0  TCP 127.0.0.1:3000->127.0.0.1:36113 (ESTABLISHED)
+node     66 root   10u  IPv4 219916      0t0  TCP 127.0.0.1:3001 (LISTEN)
+node     66 root   11u  IPv4 219917      0t0  TCP 127.0.0.1:36113->127.0.0.1:3000 (ESTABLISHED)
+node     66 root   12u  IPv4 219918      0t0  TCP 127.0.0.1:3001->127.0.0.1:43323 (ESTABLISHED)
+root:/#
+```
+
+Armed with this knowledge, we can try to make a partition.
+
+## Manually forming a partition
+
+The naïve approach to make a partition between `a` and `b` is to block incoming
+connections from and to port `3000`: then no packet will leave `a`, and we will
+have a partition. However, this misses the fact that ephemeral connections are
+used for relaying traffic between nodes, and, in this case, connection from
+`127.0.0.1:43323` (`a`) to `127.0.0.1:3001` is established and... misses the
+firewall! We could block port `3001` too, but, with more nodes, that
+would create a cluster with N partitions (N being the number of nodes) -- not
+what we want. In our example, we want two partitions.
+
+With that in mind, a bit more sophistication in firewall rules is required. To
+easily create a partition in `tick-cluster` locally, we created
+`tools/make_partitions`, which, by reading the state of the connections from
+`lsof`, will emit `iptables`/`pf` commands accordingly.
+
+In the example above, firewall rules to create a partition will look as follows
+(OS X):
+
+```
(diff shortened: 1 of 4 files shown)


docs/partitions.md
```diff
@@ -1,226 +1,59 @@
-# Partitions
+# Partition Healing
 
-In the original implementation of ringpop, if a cluster is split to multiple
-partitions, nodes in each partition declare each other as faulty, and afterward
-will no longer communicate. Ringpop implemented support for merging the
-partitions, which we call `healing`.
+In the original implementation of ringpop, if a cluster is split to multiple partitions, nodes in each partition declare each other as faulty, and afterward will no longer communicate. Ringpop implemented support for merging the partitions, which we call `healing`.
```
CorgiMan: If a cluster's **network** is split
Is more appropriate to offer distinction between network and a ringpop partition (with healed network)

CorgiMan: Maybe also add something like:
If the network partition is healed, the ringpop cluster is still partitioned into smaller ringpop clusters that don't ping each other. We introduce Partition Healing to merge these smaller clusters back into one healthy ringpop cluster.

docs/partitions.md
```diff
@@ -1,226 +1,59 @@
...
-### Start the partition
-
-First, check how the firewall rules would look like before applying them to the
-firewall (optionally, you can pass `--platform=darwin` or `--platform=linux` to
-the `make_partition` script:
-
-```shell
-$ sudo lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003
-```
+This document describes a strategy to heal a partitioned ringpop cluster. The first section describes the mechanism a node triggers when it attempts to heal a partition, the second describes how, how often and when a node should trigger this mechanism.
```
CorgiMan: I think paragraph/chapter is more appropriate

docs/partitions.md
```diff
@@ -1,226 +1,59 @@
...
 
-If you are happy with the output, apply the rules:
+### When and How Often
 
-OS X:
+Executing the algorithm from the previous section on every tick doesn't scale well. The discovery provider will be overloaded and the amount of requests grows linearly with the cluster size. To deal with this issue, we introduce a separate timer that periodically ticks on a configurable duration `T`. Every tick, there is a probability `P` of executing the partition healing algorithm.
```
CorgiMan: No mention of the discovery provider has been made yet

docs/make-partition.md
```diff
@@ -0,0 +1,194 @@
...
+```shell
+$ sudo lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003
+```
+
+If you are happy with the output, apply the rules:
+
+OS X:
+
+```shell
+$ sudo lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003 | sudo pfctl -emf -
```
mennopruijssers: I noticed that (because of the piping) when you need to enter you sudo password, this command actually fails. I used `sudo sh -c "lsof -Pnni | ./make_partition 3000-3010 3011-3024  | pfctl -emf -"` to circumvent that problem (less readable but it works ;))

docs/make-partition.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+Linux:
+
+```
+$ sudo lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003 | sudo iptables-restore
```
mennopruijssers: That's probably the same here:
sudo sh -c "lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003 | iptables-restore"

docs/make-partition.md
```diff
@@ -0,0 +1,194 @@
...
+
+On partition, `ringpop-admin top` (opened before the partition) should display
+something like this:
+
+```
+Address          P1       P2
+127.0.0.1:3000   faulty   alive
+127.0.0.1:3001   faulty   alive
+127.0.0.1:3002   alive    faulty
+127.0.0.1:3003   alive    faulty
```
mennopruijssers: Maybe we need to update this part a little bit due to reaping faulty nodes.
It will show a 2 node cluster on both partitions.

docs/make-partition.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+`ringpop-admin partitions` (opened before forming a partition) shows a more
+high-level view:
+
+```
+10:37:04.878   Checksum     # Nodes   # Alive   # Suspect   # Faulty   Sample Host
+               400620880    2         2         0           2          127.0.0.1:3002
```
mennopruijssers: Same here

docs/make-partition.md
```diff
@@ -0,0 +1,194 @@
...
+               3283514511   2         2         0           2          127.0.0.1:3000
+```
+
+That's it, we have a partition! To break it, we need to wipe the firewall rules:
+
+* OS X: `pfctl -f /etc/pf.conf`.
+* Linux: `iptables -F`.
+
+... and wait for partition healing to kick in.
```
mennopruijssers: Might be worth mentioning `ringpop-admin heal` here as well?

docs/partitions.md
```diff
@@ -1,226 +1,66 @@
...
-```
+Let `N` be the number of hosts the previous query to the discovery provider has given us. If we, for example set `P = 3/N` and `T = 30 s`, we get on average 6 heal attempts per minute in the entire cluster, with a 95% probability that there is at least one heal attempt in 30 seconds. This means that we only query the discovery provider six times per minute regardless the size of the cluster.
 
-In a few seconds, you should see output from `tick-cluster` that some of the
-nodes aren't able to ping each other. Let's verify we actuall have a partition.
+### Configurables
 
-### Checking in the tools
+* `T` -- partition healer execution interval, seconds. Configurable for application.
+* `P` -- probability of executing the partition healing algorithm. Well-displayed constant in ringpop. Easily accessible for ringpop developers.
```
mennopruijssers: These are both not (yet) configurable in ringpop-go.

docs/partitions.md
```diff
@@ -1,226 +1,66 @@
...
-`ringpop-admin partitions` (opened before forming a partition) shows a more
-high-level view:
+#### Part 1 -- making memberships compatible
 
-```
-10:37:04.878   Checksum     # Nodes   # Alive   # Suspect   # Faulty   Sample Host
-               400620880    2         2         0           2          127.0.0.1:3002
-               3283514511   2         2         0           2          127.0.0.1:3000
-```
+Goal of the first part is to make sure the membership lists of both partitions are _compatible_. That is, if they are merged according to the SWIM rules, no new faulty nodes shoul be created. To reach our goal we need to bump the incarnation numbers of nodes on both sides of the partition (because a state with a higher incarnation number always has precedence regardless of the status). Here's how we do it:
```
mennopruijssers: s/shoul/should

docs/partitions.md
```diff
@@ -1,226 +1,66 @@
...
 
-That's it, we have a partition! To break it, we need to wipe the firewall rules:
+1. `c` (coordinator) downloads `t`'s (target's, in the other partition) membership list by doing a `/join` call.
+2. `c` pings node `t` saying: the nodes that are faulty according to `c`, but alive according to `t`, are suspects. Then `t` will disseminate suspect messages to its own partition, and all nodes in `t`'s partition will reincarnate.
+3. mark all `c`'s nodes suspect, and disseminate that information in `c`'s cluster. That will trigger the `c`'s partition to reincarnate.
```
mennopruijssers: Not all of `c`'s nodes right? only they ones that are faulty according to `t`

mennopruijssers:
@Motiejus overall really good and clear writeup. Thanks for doing this!!

docs/make-partition.md
Motiejus: Can you elaborate on why it fails? It shouldn't (and doesn't fail for me).

I explicitly avoided sudo for `make_partition`, because I don't expect a user to trust a random script which executes as root, even when it shouldn't need to.

docs/make-partition.md
Motiejus: This will not work with a short reaping-faulty-node timeout! :'(

I think it's a problem.. Do you think there is an easy way to extend the faulty node timeout? I guess we still want to be able to simulate the partitions and healing..

docs/make-partition.md
Motiejus: Good catch! I`ll add it.

docs/make-partition.md
CorgiMan: I also had no problems running it like it is

mennopruijssers: Could be my shell (I'm using zsh) but when I run it, it asks twice for my sudo password but when I enter my password it says "Sorry, try again." (meaning: incorrect password while I'm sure it's the right one).

```
$ sudo -k
$ sudo lsof -Pnni | ./tools/make_partition 3000,3001 3002,3003 | sudo pfctl -emf -
Password:Password:
Sorry, try again.
Password:
Sorry, try again.
Password:
Sorry, try again.
Password:
```

docs/make-partition.md
mennopruijssers: We could change the default transition timeouts in main.js?

Current is (https://github.com/uber/ringpop-node/blob/master/main.js#L53):
```
stateTimeouts: {
            faulty: 5 * 1000, // 5s
            tombstone: 5 * 1000 // 5s
        }
```

Motiejus: Unfortunately, this will make our integration tests to run longer (i.e. at least the thombstone timeout time).

I see a few approaches here:

1. Document here which values should be increased to observe the documented behavior.
2. Change the default timeout in testpop, and make a way to do it for integration tests (command-line switch?).

I am happy to implement either. @thanodnl, what do you prefer?

mennopruijssers: I would prefer using sensible defaults in testpop and +1 for an option to speed things up for integration tests. We could go for a command-line switch but I would prefer a `--tombstone-period` flag over a `--integration-tests` flag: we might want to use the modified tombstone-period only for the tombstone tests.

Motiejus:
Now that https://github.com/uber/ringpop-node/pull/280 and https://github.com/uber/ringpop-go/pull/148 are merged, can we land this? ping  @mennopruijssers  @CorgiMan

mennopruijssers:
LGTM!

PR #61, opened 2016-05-26
Fix a flappy reaping-faulty-nodes test

Motiejus:
It is not uncommon for SUT to ping an unrelated node with itself saying
"I'm alive". In this test, we need to ignore these pings. The case is
best reproduced with an overloaded system with n > 10. Example output:

    not ok 4979 The gossip should contain a flagged

      ---
        operator: ok
        expected: true
        actual:   false
        at: Array.validateEventBody [as 8]
    op-node/test/ringpop-common/test/ringpop-assert.js:147:11)
      ...
    ============== error details ===============

    {
      "body": {
        "checksum": 3250656653,
        "changes": [
          {
            "id": "6768c914-06e1-4122-8d00-0a2648f7f86d",
            "address": "127.0.0.1:24935",
            "status": "alive",
            "incarnationNumber": 1464253153302
          }
        ],
        "source": "127.0.0.1:24935",
        "sourceIncarnationNumber": 1464253153302
      }
    }

    ============================================

### DIFF
--- a/test/reaping-faulty-nodes.js
+++ b/test/reaping-faulty-nodes.js
@@ -218,9 +218,10 @@ test2('tombstone should be gossiped with flag when applied as state', getCluster
             dsl.consumePings(t, tc),
 
             // Wait for a ping from the SUT and validate that it does not gossip about the tombstone
-            dsl.validateEventBody(t, tc, {
-                type: events.Types.Ping,
-                direction: 'request'
+            dsl.validateEventBody(t, tc, function(ping) {
+                return ping.type === events.Types.Ping
+                    && ping.direction === 'request'
+                    && ping.body.changes[0].status !== 'alive'
             }, "The gossip should contain a flagged tombstone", function (ping) {
                 return _.filter(ping.body.changes, { status: 'faulty', tombstone: true }).length === 1
                     && _.filter(ping.body.changes, { status: 'tombstone' }).length === 0;
@@ -264,9 +265,10 @@ test2('tombstone should be gossiped with flag when applied as a flag', getCluste
             dsl.consumePings(t, tc),
 
             // Wait for a ping from the SUT and validate that it does not gossip about the tombstone
-            dsl.validateEventBody(t, tc, {
-                type: events.Types.Ping,
-                direction: 'request'
+            dsl.validateEventBody(t, tc, function(ping) {
+                return ping.type === events.Types.Ping
+                    && ping.direction === 'request'
+                    && ping.body.changes[0].status !== 'alive'
             }, "The gossip should contain a flagged tombstone", function (ping) {
                 return _.filter(ping.body.changes, { status: 'faulty', tombstone: true }).length === 1
                     && _.filter(ping.body.changes, { status: 'tombstone' }).length === 0;
--- a/test/reaping-faulty-nodes.js
+++ b/test/reaping-faulty-nodes.js
@@ -222,10 +222,16 @@ test2('tombstone should be gossiped with flag when applied as state', getCluster
             // Since we pinged SUT from nodeIx=0, the node filters that
             // changes, therefore the conditional to skip it's ping request
             // below.
+            //
+            // We also ignore ping requests that gossip 'alive' status.
+            // That can happen because, at least in ringpop-node, a ping request
+            // before disseminator update could have been executed after the ping
+            // response above is sent over the wire.
             dsl.validateEventBody(t, tc, function(ping) {
                 return ping.type === events.Types.Ping
                     && ping.direction === 'request'
-                    && ping.receiver !== tc.fakeNodes[0].getHostPort()
+                    && ping.receiver !== tc.fakeNodes[0].getHostPort() // ignore nodeIx=0
+                    && ping.body.changes[0].status !== 'alive'  // ignore old ping
             }, "The gossip should contain a flagged tombstone", function (ping) {
                 return _.filter(ping.body.changes, { status: 'faulty', tombstone: true }).length === 1
                     && _.filter(ping.body.changes, { status: 'tombstone' }).length === 0;
@@ -273,10 +279,16 @@ test2('tombstone should be gossiped with flag when applied as a flag', getCluste
             // Since we pinged SUT from nodeIx=0, the node filters that
             // changes, therefore the conditional to skip it's ping request
             // below.
+            //
+            // We also ignore ping requests that gossip 'alive' status.
+            // That can happen because, at least in ringpop-node, a ping request
+            // before disseminator update could have been executed after the ping
+            // response above is sent over the wire.
             dsl.validateEventBody(t, tc, function(ping) {
                 return ping.type === events.Types.Ping
                     && ping.direction === 'request'
-                    && ping.receiver !== tc.fakeNodes[0].getHostPort()
+                    && ping.receiver !== tc.fakeNodes[0].getHostPort() // ignore nodeIx=0
+                    && ping.body.changes[0].status !== 'alive'  // ignore old ping
             }, "The gossip should contain a flagged tombstone", function (ping) {
                 return _.filter(ping.body.changes, { status: 'faulty', tombstone: true }).length === 1
                     && _.filter(ping.body.changes, { status: 'tombstone' }).length === 0;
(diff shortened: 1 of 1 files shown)


thanodnl:
Why is it failing in the first place, the line above the call to `validateEventBody` consumes all the pings that are outstanding so the ping that is validated should be a ping with the tombstone flag. What is the reason that it still has an `alive` ping received?

Motiejus:
My guess is: right after all pings are consumed, SUT pings a node with index between 2 and 9 (in case of 10-node cluster), and that gets caught in the test. The ping with tombstone comes only after the "regular" ping. That's why it is easiest to reproduce it with a higher number of nodes.

Does that make sense?

test/reaping-faulty-nodes.js
```diff
@@ -218,9 +218,10 @@ test2('tombstone should be gossiped with flag when applied as state', getCluster
             dsl.consumePings(t, tc),
 
             // Wait for a ping from the SUT and validate that it does not gossip about the tombstone
-            dsl.validateEventBody(t, tc, {
-                type: events.Types.Ping,
-                direction: 'request'
+            dsl.validateEventBody(t, tc, function(ping) {
```
thanodnl: my suspicion is that the flap happens when the ping sent is to fakeNode 0. This node is the source of the gossip and therefor the tombstone is filtered out in the ping.

thanodnl: Lets add information to all events to the index of the fakenode it has been sent to.

Motiejus:
@thanodnl correct observation regarding the source of changes. I reproduced the test again, the `alive` message comes from node 0:

```
ts=1464343723.762 session=8 started src=127.0.0.1:32446 --> dst=127.0.0.1:19416 on lo
ts=1464343723.770 session=8 127.0.0.1:32446 --> 127.0.0.1:19416 frame=1 type=0x01
INIT REQUEST id=0x0001 (1) version=2
headers
  host_port: 127.0.0.1:27649
  process_name: node[41743]
```


```
============== error details ===============

{
  "body": {
    "checksum": 2357575861,
    "changes": [
      {
        "id": "b4f6d2a5-1a54-48a2-8c0d-2505525af34f",
        "address": "127.0.0.1:19416",
        "status": "alive",
        "incarnationNumber": 1464343723403
      }
    ],
    "source": "127.0.0.1:19416",
    "sourceIncarnationNumber": 1464343723403
  }
}

============================================

============== node index : hostport ===============
sut: 127.0.0.1:19416
0 => 127.0.0.1:27649
1 => 127.0.0.1:32624
2 => 127.0.0.1:30140
3 => 127.0.0.1:45349
4 => 127.0.0.1:25301
5 => 127.0.0.1:52966
6 => 127.0.0.1:50464
7 => 127.0.0.1:24664
8 => 127.0.0.1:55776
9 => 127.0.0.1:48756
====================================================
```


```
ts=1464343723.776 session=8 127.0.0.1:32446 <-- 127.0.0.1:19416 frame=2 type=0x04 Ok
CALL RESPONSE id=0x0002 (2) flags=0x00
headers
  as: raw
tracing: spanid=1844200942,1251754723 parentid=0,0 traceid=1844200942,1251754723 flags=0x00
args[0]
    00:                                          empty
args[1]
    00:                                          empty
args[2]
    00: 7b22 6368 616e 6765 7322 3a5b 7b22 6964  {"changes":[{"id
    10: 223a 2262 3466 3664 3261 352d 3161 3534  ":"b4f6d2a5-1a54
    20: 2d34 3861 322d 3863 3064 2d32 3530 3535  -48a2-8c0d-25055
    30: 3235 6166 3334 6622 2c22 6164 6472 6573  25af34f","addres
    40: 7322 3a22 3132 372e 302e 302e 313a 3139  s":"127.0.0.1:19
    50: 3431 3622 2c22 7374 6174 7573 223a 2261  416","status":"a
    60: 6c69 7665 222c 2269 6e63 6172 6e61 7469  live","incarnati
    70: 6f6e 4e75 6d62 6572 223a 3134 3634 3334  onNumber":146434
    80: 3337 3233 3430 337d 5d7d                 3723403}]}
```

I will prepare a patch which will skip `alive` messages from `nodeIx=0`.

Motiejus:
@thanodnl mentioned that it would be useful to have `receiverIx` next to `receiver`, but since we have a nice convenience function `fakeNode.getHostPort()`, I don't think that's necessary (previous version had host and port concatenation). I also don't love the semantics of adding `-1` if node is not found, so I would prefer to keep it this way.

thanodnl:
LGTM

CorgiMan:
Looks good after you add that comment :)

test/reaping-faulty-nodes.js
```diff
@@ -217,7 +217,10 @@ test2('tombstone should be gossiped with flag when applied as state', getCluster
             // clear all pings that happened before we gossiped the tombstone
             dsl.consumePings(t, tc),
 
-            // Wait for a ping from the SUT and validate that it does not gossip about the tombstone
+            // Wait for a ping from the SUT and validate that it does not gossip about the tombstone.
+            //
+            // Since we pinged nodeIx=0, the node filters that changes, therefore the conditional
```
CorgiMan: We pinged the SUT from nodeIx=0 right? I don't think we pinged nodeIx=0

Motiejus: Yeah you're right. I should have been a bit slower... Updating.

Motiejus:
@CorgiMan updated.

Motiejus:
So the test is still failing if there is even more load on the system. Here's the test log:

```
./../ringpop-node/main.js --listen=127.0.0.1:27458 --hosts=/tmp/ringpop-integration-test-hosts.json
DEBUG: Throttling 62926 (node)
* starting requestAdminStats
* starting waitForStatsAssertDetectChecksumMethod
* starting waitForJoins
* starting requestAdminStats
* starting waitForStatsAssertStatus
* starting sendPing
* starting waitForPingResponse
* starting consumeRequests
* starting validateEventBody
not ok 43 The gossip should contain a flagged tombstone
  ---
    operator: ok
    expected: true
    actual:   false
    at: Array.validateEventBody [as 8] (/Users/motiejus/Uber/ringpop-common/test/ringpop-assert.js:147:11)
  ...
============== error details ===============

{
  "body": {
    "checksum": 3511615293,
    "changes": [
      {
        "id": "9e2248b2-048a-4787-88e1-fa7cff365446",
        "address": "127.0.0.1:27458",
        "status": "alive",
        "incarnationNumber": 1464699381511
      }
    ],
    "source": "127.0.0.1:27458",
    "sourceIncarnationNumber": 1464699381511
  }
}

============================================

* starting expectOnlyPingsAndPingReqs
# cluster-size 21: tombstone should be gossiped with flag when applied as a flag
./../ringpop-node/main.js --listen=127.0.0.1:15554 --hosts=/tmp/ringpop-integration-test-hosts.json
DEBUG: Throttling 63189 (node)
* starting requestAdminStats
* starting waitForStatsAssertDetectChecksumMethod
* starting waitForJoins
* starting requestAdminStats
* starting waitForStatsAssertStatus
* starting sendPing
* starting waitForPingResponse
* starting consumeRequests
* starting validateEventBody
* starting expectOnlyPingsAndPingReqs

1..90
# tests 90
# pass  89
# fail  1
```

log of the session here (view with `less -R`): [tcap.txt](https://github.com/uber/ringpop-common/files/291119/tcap.txt)

mennopruijssers:
LGTM!

CorgiMan:
Looks good!

It seems that all the problems in this pr would've been solved if we would send a ping, instead of waiting for one from the sut. Something to keep in mind for the future.

test/reaping-faulty-nodes.js
```diff
@@ -217,10 +217,21 @@ test2('tombstone should be gossiped with flag when applied as state', getCluster
             // clear all pings that happened before we gossiped the tombstone
             dsl.consumePings(t, tc),
 
-            // Wait for a ping from the SUT and validate that it does not gossip about the tombstone
-            dsl.validateEventBody(t, tc, {
-                type: events.Types.Ping,
-                direction: 'request'
+            // Wait for a ping from the SUT and validate that it does not gossip about the tombstone.
```
thanodnl: Even though this line came from the existing code, when I read it again here it seemed a bit misleading. Because we are making sure it _does_ gossip about the tombstone, but we make sure it does not gossip the tombstone state. If you can add `state` or `as a state` to the end of the line it does make more sense.

Motiejus: Good remark. I refactored the checking part to a separate function and updated the comment.

PR #63, opened 2016-05-27
Update tick-cluster to be in sync with ringpop-node version

btromanova:
There were some changes added to ringpop-node version of tick-cluster, which is used in ringpop-admin tests. We want to have only one version of tick-cluster that should leave in ringpop-common. This PR is to put ringpop-common version in sync with ringpop-node.

### DIFF
--- a/tools/tick-cluster.js
+++ b/tools/tick-cluster.js
@@ -243,6 +243,11 @@ function debugClear() {
     });
 }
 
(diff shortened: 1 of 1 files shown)


dansimau:
LGTM

jwolski:
lgtm

Motiejus:
LGTM

PR #64, opened 2016-05-30
Add arguments for alive, suspect and faulty periods

CorgiMan:
Run the test program with the following flags: suspect period of 5 seconds, faulty period of 5 seconds and tombstone period of 5 seconds.

### DIFF
--- a/test/test-coordinator.js
+++ b/test/test-coordinator.js
@@ -165,11 +165,27 @@ TestCoordinator.prototype.startSUT = function startSUT() {
     var newProc;
     var hostsFileArg = util.format('--hosts=%s', this.hostsFile);
     var listenArg = util.format('--listen=%s', this.sutHostPort);
+    var suspectPeriodArg = util.format('--suspect=5')
+    var faultyPeriodArg = util.format('--faulty=5')
+    var tombstonePeriodArg = util.format('--tombstone=5')
     console.log(this.sutProgram, listenArg, hostsFileArg);
     if (this.sutInterpreter) {
-        newProc = childProc.spawn(this.sutInterpreter, [this.sutProgram, listenArg, hostsFileArg]);
+        newProc = childProc.spawn(this.sutInterpreter, [
+            this.sutProgram,
+            listenArg,
+            hostsFileArg,
+            suspectPeriodArg,
+            faultyPeriodArg,
+            tombstonePeriodArg,
+        ]);
     } else {
-        newProc = childProc.spawn(this.sutProgram, [listenArg, hostsFileArg]);
+        newProc = childProc.spawn(this.sutProgram, [
+            listenArg,
+            hostsFileArg,
+            suspectPeriodArg,
+            faultyPeriodArg,
+            tombstonePeriodArg,
+        ]);
     }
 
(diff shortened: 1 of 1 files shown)


dansimau:
typo in commit message: s/argements/arguments!

test/test-coordinator.js
```diff
@@ -165,11 +165,27 @@ TestCoordinator.prototype.startSUT = function startSUT() {
     var newProc;
     var hostsFileArg = util.format('--hosts=%s', this.hostsFile);
     var listenArg = util.format('--listen=%s', this.sutHostPort);
+    var suspectPeriodArg = util.format('--suspect=5')
```
mennopruijssers: No need for util.format since those are static strings.

test/test-coordinator.js
```diff
@@ -165,11 +165,27 @@ TestCoordinator.prototype.startSUT = function startSUT() {
...
+    var faultyPeriodArg = util.format('--faulty=5')
+    var tombstonePeriodArg = util.format('--tombstone=5')
     console.log(this.sutProgram, listenArg, hostsFileArg);
     if (this.sutInterpreter) {
```
mennopruijssers: If we add more arguments it might be better to refactor this bit a bit. Something like:
```lang=javascript
var program;
var arguments = [];
if (this.sutInterpreter) {
    program = this.sutInterpreter;
    arguments.push(this.sutProgram);
} else {
    program = this.sutProgram;
}
var arguments = [];
arguments.push(util.format('--hosts=%s', this.hostsFile));
arguments.push(util.format('--listen=%s', this.sutHostPort);
arugments.push(util.format('--suspect=5'));
// etc. other arguments.

newProc = childProc.spawn(program, arguments);
```

test/test-coordinator.js
```diff
@@ -163,14 +166,22 @@ TestCoordinator.prototype.start = function start(callback) {
...
+        program = this.sutInterpreter
+        args.push(this.sutProgram)
     }
+    args.push(util.format('--hosts=%s', this.hostsFile));
+    args.push(util.format('--listen=%s', this.sutHostPort));
+    args.push(util.format('--suspect-period=%d', this.suspectPeriod));
+    args.push(util.format('--faulty-period=%d', this.faultyPeriod));
+    args.push(util.format('--tombstone-period=%d', this.tombstonePeriod));
+
+    console.log(program, args.join(" "));
```
Motiejus: I would rather do something like:

```
+    console.log("Starting '" + program + " " + args.join(" ") + "'");
```

Also, a colon in `console.log` behaves weirdly in my experience -- what does the actual log output look like?

CorgiMan: At the start of every test this is putout
```
# cluster-size 5: bidirectional full sync test
./main.js --hosts=/tmp/ringpop-integration-test-hosts.json --listen=127.0.0.1:24701 --suspect-period=5000 --faulty-period=5000 --tombstone-period=5000
```

I think it's clear enought due to the `./`

mennopruijssers:
LGTM

Motiejus:
LGTM

PR #65, opened 2016-05-31
kill and suspend arbitrary number of processes in tick-cluster.

btromanova:
This pr should solve two issues:
- currently if we try to kill or suspend more processes number of 'alive', tick-cluster will get into infinite loop.
- currently it's not possible to kill or suspend more then 9 processes.

### DIFF
--- a/tools/tick-cluster.js
+++ b/tools/tick-cluster.js
@@ -21,6 +21,7 @@
 
 'use strict';
 
+var _ = require('lodash');
 var childProc = require('child_process');
 var color = require('cli-color');
 var farmhash = require('farmhash').hash32;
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "lodash": "^4.6.1",
     "strformat": "^0.0.7",
     "uber-licence": "^2.0.1",
+    "underscore": "^1.5.2",
```
Motiejus: It feels strange to have both underscore and lodash in dependencies. Can we get rid of one of them?

mennopruijssers:
Now you need always to hit enter after a command. Is it possible to only wait for the enter when inputting a digit? 
Also, `space` to draw a horizontal line doesn't work anymore (probably because of the trim)

btromanova:
@mennopruijssers Switched rawInputMode back to true, now there is no need for enter, but still possible to insert multidigit number of processes to kill or suspend.

tools/tick-cluster.js
```diff
@@ -309,18 +311,21 @@ function onData(char) {
...
             state = 'top';
             func = null;
+            num = 0;
         } else {
             console.error('expecting: ' + numArgRe);
             state = 'top';
+            num = 0;
         }
     } else if (state === 'readchar') {
         if (char.match(debugRe)) {
```
mennopruijssers: It might be nice to write the char here to stdout as well.

mennopruijssers:
LGTM!

Motiejus:
LGTM

PR #68, opened 2016-06-07
Add controls to tick cluster to stop the gossip in the cluster.

thanodnl:
Tick cluster already had controls to start the gossip of the cluster, with this feature it is now also able to stop the gossiping of the cluster.

This was useful in debugging dissemination behavior step by step in combination with the tick command in tick cluster.

### DIFF
--- a/tools/tick-cluster.js
+++ b/tools/tick-cluster.js
@@ -283,10 +283,10 @@ function onData(char) {
             case 'j':
                 joinAll();
                 break;
-            case 'G':
+            case 'g':
                 stopGossip();
                 break;
-            case 'g':
+            case 'G':
                 startGossip();
                 break;
             case 's':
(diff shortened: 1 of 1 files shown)


tools/tick-cluster.js
```diff
@@ -517,6 +536,7 @@ function main() {
 function displayMenu(logFn) {
     logFn('\td <flag>\tSet debug flag');
     logFn('\tD\t\tClear debug flags');
+    logFn('\tG\t\tStop gossip');
```
mennopruijssers: the other commands are listing the lowercase option before the uppercase one. Can you follow that style here?

mennopruijssers:
LGTM!

tools/tick-cluster.js
thanodnl: Yes, maybe we should change the casing. Because normally the action is initiated with the lower case and after that reverted with the upper case. What do you think about following that paradigm as well?

mennopruijssers: totally agree. lets swap it :)

btromanova:
Can you please also update the  logMsg on line 530? After that LGTM.

thanodnl:
Ah, thanks, didn't see that one.

PR #69, opened 2016-06-07
Validate ping response after a reincarnation to be correct

thanodnl:
Ringpop used to gossip a reincarnation as coming from the node that made the initial declaration of a new state which caused the reincarnation to be filtered out to the node that made the initial declaration. Therefor delaying and potentially causing a full sync to happen when it should not have happened.

This test is to explicitly test the ping response of a ping triggering a reincarnation.

### DIFF
--- a/test/it-tests.js
+++ b/test/it-tests.js
@@ -52,7 +52,8 @@ var features = {
     'reincarnate': {
         mandatory: true,
         tests: [
-            './incarnation-no-tests'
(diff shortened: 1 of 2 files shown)


mennopruijssers:
Nice: LGTM! this problems fails until the two PR's in node and go are merged to master right?

Motiejus:
LGTM

PR #70, opened 2016-06-08
Remove redundant tests

Motiejus:
A little bit of history: a while ago, /join handler used to accept the
joining node to the membership. It is not the case anymore, so the tests
below do not really make a lot of sense. Deleting.

They are also flapping! There is a race between disableNode and
enableNode. So let's keep ourselves clean.


mennopruijssers:
LGTM!

CorgiMan:
Looks good!

PR #72, opened 2016-06-09
Fix double reincarnation

thanodnl:
This PR adds an integration tests assessing that reincarnation only happens for changes that would actually override the state of the node and don't reincarnate if the change will not result in an overridden state.

### DIFF
--- a/test/double-reincarnation.js
+++ b/test/double-reincarnation.js
@@ -1,0 +1,40 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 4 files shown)


test/double-reincarnation.js
```diff
@@ -0,0 +1,40 @@
...
+test2('ringpop should not bump incarnation number when gossip is older', getClusterSizes(2), 20000,
+    prepareCluster(function(t, tc, n) { return [
+        // do not disable node
+        dsl.sendPing(t, tc, 0, {
+            sourceIx: 0,
+            subjectIx: 'sut',
+            status: 'suspect',
+            subjectIncNoDelta: -1 // send a gossip with an older incarnation number
+        }),
+        dsl.waitForPingResponse(t, tc, 0, 1, true),
```
mennopruijssers: Why the extra 2 arguments (`1, true`)? `waitForPingResponse` only expects three? (https://github.com/uber/ringpop-common/blob/5cc7e1dcdf5fa7feb8b8b625f15ac22b5ce781f9/test/ringpop-assert.js#L269)

mennopruijssers:
one question but: LGTM!

test/double-reincarnation.js
thanodnl: Ah, that is strange. I cheated a bit and took the base of this test from here: https://github.com/uber/ringpop-common/blob/fix/double-reincarnation/test/incarnation-no-tests.js#L34

I will remove it in this test 😃

test/ringpop-assert.js
```diff
@@ -896,7 +896,7 @@ function piggyback(tc, opts) {
 
     if(opts.subjectIx === 'sut') {
         update.address = tc.sutHostPort;
-        update.sourceIncarnationNumber = tc.test_state['sutIncarnationNumber'];
```
Motiejus: Can you help me understand why this change is needed?

Motiejus:
LGTM for the test, but see my question about the last line of the patch.

test/ringpop-assert.js
thanodnl: It is hard to see without more context, but the `piggyback` function consists of roughly two parts. The first part is about the source of the gossip, the second part is about the subject. We are here in the part that is about the subject.

Even though we are setting the subject information this line had a bug where the incarnation number of the source of the gossip was set to the incarnation number of the sut. If you look at a bit more context below this line you will see that all branches set the `incarnationNumber` except for this branch.

Motiejus: That makes sense. Thank you!

Motiejus:
LGTM

PR #73, opened 2016-06-16
static stat analysis

CorgiMan:
Part 1 for the cluster-test-orchestrator. One file contains the logic to scan a window of a file containing ringpop stats inserted with labels that indicate when the commands of a script are run. The other file contains analysis for: convergence time; number of converged checksums; counting of individual stats.

For reference, here is a small sample of what such a stat file would look like.

```
2016-06-16T16:22:57.795666563Z|ringpop.172_18_24_214_3000.make-alive:1|c
2016-06-16T16:22:57.795735453Z|ringpop.172_18_24_214_3000.membership-update.alive:1|c
2016-06-16T16:22:57.795883314Z|ringpop.172_18_24_214_3000.compute-checksum:0.016676|ms
2016-06-16T16:22:57.795897982Z|ringpop.172_18_24_214_3000.checksum:1565761574|g
2016-06-16T16:22:57.795919548Z|ringpop.172_18_24_214_3000.membership.checksum-computed:1|c
2016-06-16T16:22:57.795931518Z|ringpop.172_18_24_214_3000.changes.apply:1|g
2016-06-16T16:22:57.796057927Z|ringpop.172_18_24_214_3000.ring.checksum-computed:1|c
2016-06-16T16:22:57.796070776Z|ringpop.172_18_24_214_3000.ring.checksum:3321709178|g
2016-06-16T16:22:57.796082369Z|ringpop.172_18_24_214_3000.ring.server-added:1|c
2016-06-16T16:22:57.796091444Z|ringpop.172_18_24_214_3000.ring.server-removed:0|c
2016-06-16T16:22:57.796109218Z|ringpop.172_18_24_214_3000.ring.changed:1|c
2016-06-16T16:22:57.796119814Z|ringpop.172_18_24_214_3000.membership-set.alive:1|c
2016-06-16T16:22:57.796256752Z|ringpop.172_18_24_214_3000.updates:1|c
2016-06-16T16:22:57.796293995Z|ringpop.172_18_24_214_3000.max-piggyback:0|g
```

PR #74, opened 2016-06-16
implementation of stat ingester

CorgiMan:
Part 2 of the cluster-test-orchestrator.

In this PR you find the stat ingester of the orchestrator. The stat ingester does real-time analysis to find out if the ringpop cluster is stable and writes the stats to a file for later analysis.

### DIFF
--- a/cluster-test/test-orchestrator/stat_ingestor.go
+++ b/cluster-test/test-orchestrator/stat_ingestor.go
@@ -1,0 +1,223 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


cluster-test/test-orchestrator/stat_ingestor.go
```diff
@@ -0,0 +1,223 @@
...
+		hs := strings.Replace(h, ".", "_", -1)
+		hs = strings.Replace(hs, ":", "_", -1)
+		if empty, ok := si.emptyNodes[hs]; !ok || !empty {
+			return false
+		}
+	}
+	return true
+}
+
+// InsertLabel writes a label into the stats file.
```
Motiejus: Like mentioned in #73, it would be awesome to have more detail on what labels are and what they are for.

cluster-test/test-orchestrator/stat_ingestor.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/pkg/errors"
+)
+
+// The StatIngester is a UDP server that accepts ringpop stats with added
+// timestamps. The StatIngester analyzes the stream so that it knows when the
+// cluster reaches a stable state. It also writes the stream into a file for
+// later analysis.
+type StatIngester struct {
+	sync.Mutex
+	emptyNodes  map[string]bool
```
Motiejus: Can you explain how are you using this variable?

I see there is a non-trivial state machine here while parsing the stats, and this record is quite involved into it. Could you explain how that works?

cluster-test/test-orchestrator/stat_ingestor.go
```diff
@@ -0,0 +1,223 @@
...
+				return
+			}
+
+			// write to file
+			err = writeln(si.file, buf[0:n])
+			if err != nil {
+				log.Fatalln(err)
+			}
+
+			statsQueue <- []byte(string(buf[0:n]))
```
Motiejus: Should we split the UDP code and the stats-parsing state machine to separate files/modules with a clear interface between the two?

I see two benefits:
1. Easier to add other listeners (e.g. stats replay?).
2. More modular, easier to understand code.

Motiejus:
To sum up, my thoughts on this:

* Please document the stat parsing state machine, using the names of the data structures in the code.
* Split out UDP server to a separate package, and interact with the stat-parser through a public package API.
* Add tests for the state machine. Depending on how big is the UDP server and its API, maybe the server too (but if it's only one function with just a few lines, maybe not).

CorgiMan:
Now #78

PR #75, opened 2016-06-16
implementation of measurements and assertions

CorgiMan:
Part 3 of cluster-test-orchestrator

This PR contains the implementation of the measurement and assertion structs.

The Measurement struct has a method called Measure which makes calls to the static analysis function in PR #73

### DIFF
--- a/cluster-test/test-orchestrator/assertion.go
+++ b/cluster-test/test-orchestrator/assertion.go
@@ -1,0 +1,135 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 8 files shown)


cluster-test/test-orchestrator/assertion.go
```diff
@@ -0,0 +1,135 @@
...
+type Assertion struct {
+	Type AssertionType // can be is or in
+
+	// This is the Value of the Assertion in case of AssertionTypeIs
+	// or the first Value of the interval in case of AssertionTypeIn.
+	V1 Value
+
+	// The second Value of the interval in case of AssertionTypeIn.
+	// This value is ignored in case of AssertionTypeIs.
+	V2 Value
```
mennopruijssers: How extensible is this struct? will you just add `V..`'s when you need more params?

CorgiMan: Why would this struct ever need more params?

I can't think of anything we want to assert on that needs more values than an interval.

CorgiMan:
Now at #79

PR #76, opened 2016-06-16
placeholders for command, scenario structs and a calculator utility.

CorgiMan:
Part 4 of the cluster-test-orchestrator

This PR contains placeholder structs for Command and Scenario which can be filled in by the parser which will be added to Part 5 of the cluster-test-orchestrater in #77.

This PR also contains a calculator utility that is also used while parsing the test yaml-files.

### DIFF
--- a/cluster-test/test-orchestrator/calc.go
+++ b/cluster-test/test-orchestrator/calc.go
@@ -47,7 +47,8 @@ func Eval(expression string) (f float64, err error) {
 	// parse expression
 	expr, err := parser.ParseExpr(expression)
 	if err != nil {
-		return 0, errors.Wrapf(err, "eval %s\n", expression)
+		msg := fmt.Sprintf("eval error for expression: \"%s\"", expression)
+		return 0, errors.New(msg)
 	}
 
 	// evaluate expression
--- a/cluster-test/test-orchestrator/calc.go
+++ b/cluster-test/test-orchestrator/calc.go
@@ -1,0 +1,92 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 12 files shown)


cluster-test/test-orchestrator/calc.go
```diff
@@ -0,0 +1,97 @@
...
+	"go/ast"
+	"go/parser"
+	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+// Eval evaluates the value of an expression to a float64. It can be used
+// as a simple calculator. e.g: `"2+3*4" -> 14.0`.
+func Eval(expression string) (f float64, err error) {
```
mennopruijssers: I really think this class should be unit tested.

cluster-test/test-orchestrator/command.go
```diff
@@ -0,0 +1,52 @@
...
+	Label string
+
+	// Cmd can be one of:
+	// - `cluster-kill`
+	// - `cluster-start`
+	// - `cluster-rolling-restart`
+	// - `network-drop <SPLIT> <PERCENTAGE>`
+	// - `network-delay <SPLIT> <DURATION>`
+	// - `wait-for-stable`
+	Cmd string
```
mennopruijssers: If it can by one of those, why not add them as an enum?

mennopruijssers:
Looks good man but I would like to see unit tests for the calculator.

cluster-test/test-orchestrator/calc.go
CorgiMan: Tests are added

cluster-test/test-orchestrator/command.go
CorgiMan: Also, this just holds the information after the parsing (other pr) so it doesn't contain information if the commands that are passed make sense or have the correct number of arguments.

mennopruijssers:
LGTM

CorgiMan:
Now at #80

PR #77, opened 2016-06-16
parser that parsed test scenarios describe in a yaml format

CorgiMan:
Part 5 of the cluster-test-orchestrator.

This file parses the tests from a yaml format to the Scenario struct. It uses the structures defined in #75 (Measurement and Assertion) and #76 (Scenario and Command).


CorgiMan:
Now at #81

PR #78, opened 2016-06-17
Ws stat ingestor

CorgiMan:
Part 2 of the cluster-test-orchestrator.

In this PR you find the stat ingester of the orchestrator. The stat ingester does real-time analysis to find out if the ringpop cluster is stable and writes the stats to a file for later analysis.


cluster-test/test-orchestrator/stat_ingester.go
```diff
@@ -0,0 +1,175 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// This file contains is responsible for ingesting the ringpop stats of the
```
Motiejus: 👍

Motiejus:
Good work. LGTM!

cluster-test/test-orchestrator/udp_scanner.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	return &UDPScanner{
+		buf:   make([]byte, 1024),
+		sConn: sConn,
+	}, nil
+}
+
+// Scans the next line, and returns whether there is one.
+func (s *UDPScanner) Scan() bool {
```
mennopruijssers: Did you test if this is really reading a line and not the full buffer size? I checked the documentation and it doesn't like it's defined.

cluster-test/test-orchestrator/stat_ingester.go
```diff
@@ -0,0 +1,175 @@
...
+// IsClusterStable indicates, judging from the processed stats, whether the
+// cluster is in a stable state. The input are the hosts that should be
+// alive.
+func (si *StatIngester) IsClusterStable(hosts []string) bool {
+	si.Lock()
+	defer si.Unlock()
+
+	for _, h := range hosts {
+		hs := strings.Replace(h, ".", "_", -1)
+		hs = strings.Replace(hs, ":", "_", -1)
```
mennopruijssers: you could combine those two replace methods by using a regex:
```
re := regexp.MustCompile("[.:]")
hs := re.ReplaceAllString("h", "_")
```
You could even make the regex a global var so you don't have to initialize it every time :)

cluster-test/test-orchestrator/stat_ingester.go
```diff
@@ -0,0 +1,175 @@
...
+
+	return nil
+}
+
+// InsertLabel writes a line like "label:t0|cmd: kill 1" into the stats file.
+// The line indicates at what time a command is run. The idea is that all stats
+// that are recorded between two labels can be used to measure the effect of
+// the command associated with the first label.
+func (si *StatIngester) InsertLabel(label, cmd string) {
+	fmt.Fprintf(si.writer, "label:%s|cmd: %s\n", label, cmd)
```
mennopruijssers: Shouldn't you serialize access to the file with a mutex/lock as well?

cluster-test/test-orchestrator/stat_ingester.go
```diff
@@ -0,0 +1,175 @@
...
+	// disseminated by any node, the cluster is said to be stable.
+	// emptyNodes holds track of which nodes are empty and which nodes still
+	// have changes to disseminate.
+	emptyNodes map[string]bool
+
+	// When waiting for the cluster to be stable, we first want to make sure
+	// that the cluster was unstable at some point. This makes sure that any
+	// failure condition we throw at the cluster has taken effect before we
+	// move onto the next failure condition.
+	wasUnstable bool
```
mennopruijssers: Does a name like 'unstable' or 'isUnstable' make more sense? This bool describes the state of the cluster as it currently is observed right?

cluster-test/test-orchestrator/stat_ingester_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func ExampleWaitForStable() {
+	si := NewStatIngester(nopWriter{})
+	scanner := bufio.NewScanner(strings.NewReader(stats2))
+	si.IngestStats(scanner)
+	si.WaitForStable(
+		[]string{"172.18.24.220:3000", "172.18.24.220:3001"},
+	)
+
+	// Output:
```
mennopruijssers: This example isn't "testing" anything right? Is it possibly to make it a unit test?

cluster-test/test-orchestrator/stat_ingester.go
```diff
@@ -0,0 +1,175 @@
...
+// waits for the cluster to first become unstable if it isn't already, and then
+// blocks until the cluster has reached a stable state again.
+func (si *StatIngester) WaitForStable(hosts []string) {
+	// wait for cluster to become unstable
+	for !si.wasUnstable {
+		time.Sleep(200 * time.Millisecond)
+	}
+	// wait for cluster to become stable
+	for !si.IsClusterStable(hosts) {
+		time.Sleep(200 * time.Millisecond)
```
mennopruijssers: Is it possible to use a channel instead of a sleep for it?

PR #80, opened 2016-06-17
Ws structs for parsing

CorgiMan:
Part 4 of the cluster-test-orchestrator

This PR contains placeholder structs for Command and Scenario which can be filled in by the parser which will be added to Part 5 of the cluster-test-orchestrater in #81 .

This PR also contains a calculator utility that is also used while parsing the test yaml-files.


CorgiMan:
This one already got a LGTM from @mennopruijssers in #76

PR #81, opened 2016-06-17
parser that parsed test scenarios describe in a yaml format

CorgiMan:
Part 5 of the cluster-test-orchestrator.

This file parses the tests from a yaml format to the Scenario struct. It uses the structures defined in #75 (Measurement and Assertion) and #76 (Scenario and Command).

### DIFF
--- a/cluster-test/test-orchestrator/test_yaml_parser.go
+++ b/cluster-test/test-orchestrator/test_yaml_parser.go
@@ -21,7 +21,6 @@
 package main
 
 import (
-	"bufio"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -79,6 +78,11 @@ func parseScenarios(bts []byte) []*Scenario {
 func extractScenarios(runs *testYaml) []*Scenario {
 	var result []*Scenario
 	for _, scenarioData := range runs.Scenarios {
+		for _, vari := range scenarioData.Runs[0] {
+			if vari[0] != '<' || vari[len(vari)-1] != '>' {
+				panic(fmt.Sprintf("variable '%s' not of the from <var>", vari))
+			}
+		}
 		// We start at i=1 because the first entry of the runs declares the
 		// variables names. e.g. [<N>, <A>, <B>].
 		for i := 1; i < len(scenarioData.Runs); i++ {
@@ -122,7 +126,7 @@ func extractScenario(data *scenarioYaml, runIx int) *Scenario {
 	for i := range data.Measure {
 		measureStrs[i] = replace(data.Measure[i], varsData, runData)
 	}
-	measure := parseMeasure(measureStrs)
+	measure := parseMeasurements(measureStrs)
 
 	return &Scenario{
 		Name:    name,
@@ -166,19 +170,19 @@ func parseScript(labels, cmdStrs []string) []*Command {
 
 func parseCommand(label, cmdString string) *Command {
 	defer wrapPanicf("in parse command '%s: %s'", label, cmdString)
-	split := split(cmdString)
-	if len(split) == 0 {
+	fields := strings.Fields(cmdString)
+	if len(fields) == 0 {
 		panic("empty command")
 	}
 
 	return &Command{
 		Label: label,
-		Cmd:   split[0],
-		Args:  split[1:],
+		Cmd:   fields[0],
+		Args:  fields[1:],
 	}
 }
 
-func parseMeasure(msData []string) []*Measurement {
+func parseMeasurements(msData []string) []*Measurement {
 	var ms []*Measurement
 	for _, mData := range msData {
 		ms = append(ms, parseMeasurement(mData))
@@ -189,36 +193,36 @@ func parseMeasure(msData []string) []*Measurement {
 func parseMeasurement(str string) *Measurement {
 	defer wrapPanicf("in parse measure '%s'", str)
 
-	split := split(str)
-	var measurementArgs = split
-
-	if len(split) < 3 {
+	fields := strings.Fields(str)
+	if len(fields) < 3 {
 		panic("contains too few fields")
 	}
 
-	// search for assertion
+	measurementArgs := fields[3:]
+
+	// search for optional assertion
 	var assertion *Assertion
-	for i, s := range split {
+	for i, s := range measurementArgs {
 		if s == "is" || s == "in" {
-			interval := strings.Join(split[i+1:], "")
-			assertion = parseAssertion(split[i], interval)
-			measurementArgs = split[3:i]
+			interval := strings.Join(measurementArgs[i+1:], "")
+			assertion = parseAssertion(s, interval)
+			measurementArgs = measurementArgs[:i]
 		}
 	}
 
 	return &Measurement{
-		Start:     split[0],
-		End:       split[1],
-		Quantity:  split[2],
+		Start:     fields[0],
+		End:       fields[1],
+		Quantity:  fields[2],
 		Args:      measurementArgs,
 		Assertion: assertion,
 	}
 }
 
-func parseAssertion(typStr string, arg string) *Assertion {
-	defer wrapPanicf("in parse assertion '%s %s'", typStr, arg)
+func parseAssertion(typeStr string, arg string) *Assertion {
+	defer wrapPanicf("in parse assertion '%s %s'", typeStr, arg)
 
-	switch typStr {
+	switch typeStr {
 	case "is":
 		typ := AssertionTypeIs
 		v := parseValue(arg)
@@ -264,7 +268,7 @@ func parseRange(rng string) (v1, v2 Value) {
 func parseValue(str string) Value {
 	defer wrapPanicf("in parse value '%s", str)
 
-	// First check if the input is a numer or expression.
+	// First check if the input is a number or expression.
 	v, err := Eval(str)
 	if err == nil {
 		return v
@@ -278,24 +282,6 @@ func parseValue(str string) Value {
 	}
 
 	panic("value is not a number duration or expression")
-}
-
-// split splits a whitespace separted string
-func split(str string) []string {
-	defer wrapPanicf("in split '%s'", str)
-
-	scanner := bufio.NewScanner(strings.NewReader(str))
-	scanner.Split(bufio.ScanWords)
-
-	var wrds []string
-	for scanner.Scan() {
-		wrds = append(wrds, scanner.Text())
-	}
-	if err := scanner.Err(); err != nil {
-		panic(err)
-	}
-
-	return wrds
 }
 
 // replace finds occurrences of varsData and replaces them by the respective
--- a/cluster-test/test-orchestrator/test_yaml_parser.go
+++ b/cluster-test/test-orchestrator/test_yaml_parser.go
@@ -138,6 +138,8 @@ func extractScenario(data *scenarioYaml, runIx int) *Scenario {
 }
 
 func extractScript(script []map[string]string, varsData, runData []string) (labels, cmds []string) {
+	labels = make([]string, 0, len(script))
+	cmds = make([]string, 0, len(script))
 	for _, cmdData := range script {
 		if len(cmdData) != 1 {
 			// We are asserting that commands are one line only to comply with
@@ -146,7 +148,7 @@ func extractScript(script []map[string]string, varsData, runData []string) (labe
 			// script:
 			// - t0: command1
 			// - t1: command2
-			panic(fmt.Sprintf("command should contain exactly one entry, has %v", cmdData))
+			panic(fmt.Sprintf("command '%v' should contain exactly one entry", cmdData))
 		}
 
 		for label, cmd := range cmdData {
--- a/cluster-test/test-orchestrator/test_yaml_parser.go
+++ b/cluster-test/test-orchestrator/test_yaml_parser.go
@@ -79,6 +79,8 @@ func parseScenarios(bts []byte) []*Scenario {
 func extractScenarios(runs *testYaml) []*Scenario {
 	var result []*Scenario
 	for _, scenarioData := range runs.Scenarios {
+		// We start at i=1 because the first entry of the runs declares the
+		// variables names. e.g. [<N>, <A>, <B>].
 		for i := 1; i < len(scenarioData.Runs); i++ {
 			s := extractScenario(scenarioData, i)
 			result = append(result, s)
@@ -116,9 +118,9 @@ func extractScenario(data *scenarioYaml, runIx int) *Scenario {
 	script := parseScript(labels, cmds)
 
 	// extract Measure
-	var measureStrs []string
-	for _, measureStr := range data.Measure {
-		measureStrs = append(measureStrs, replace(measureStr, varsData, runData))
+	measureStrs := make([]string, len(data.Measure))
+	for i := range data.Measure {
+		measureStrs[i] = replace(data.Measure[i], varsData, runData)
 	}
 	measure := parseMeasure(measureStrs)
 
@@ -134,8 +136,13 @@ func extractScenario(data *scenarioYaml, runIx int) *Scenario {
 func extractScript(script []map[string]string, varsData, runData []string) (labels, cmds []string) {
 	for _, cmdData := range script {
 		if len(cmdData) != 1 {
-			msg := fmt.Sprintf("\"%v\" is not a valid command, should contain exactly one entry", cmdData)
-			panic(msg)
+			// We are asserting that commands are one line only to comply with
+			// the yaml that looks like:
+			//
+			// script:
+			// - t0: command1
+			// - t1: command2
+			panic("command should contain exactly one entry")
 		}
 
 		for label, cmd := range cmdData {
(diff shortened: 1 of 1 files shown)


cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+	Name string
+	Size string
+	Desc string
+
+	Script  []map[string]string
+	Measure []string
+	Runs    [][]string
+}
+
+func parse(bts []byte) (scns []*Scenario, err error) {
```
mennopruijssers: Might be misleading that the parse-method only returns scenario's instead of a testYaml-struct (with config and scenario's)

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	return extractScenarios(testYaml)
+}
+
+// extractScenarios returns a scenario for every element in the runs list.
+func extractScenarios(runs *testYaml) []*Scenario {
+	var result []*Scenario
+	for _, scenarioData := range runs.Scenarios {
+		for i := 1; i < len(scenarioData.Runs); i++ {
```
mennopruijssers: You'r missing one iteration here because you're looping from 1 till the number of runs.

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	return wrds
+}
+
+// replace finds occurrences of varsData and replaces them by the respective
+// element in the runsData.
+func replace(str string, varsData []string, runData []string) string {
+	for i := range varsData {
+		str = strings.Replace(str, varsData[i], runData[i], -1)
```
mennopruijssers: We're not prefixing vars with a special character? (e.g. a dollar sign?)

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+	if err != nil {
+		panic("size convert: " + err.Error())
+	}
+
+	// extract script
+	labels, cmds := extractScript(data.Script, varsData, runData)
+	script := parseScript(labels, cmds)
+
+	// extract Measure
+	var measureStrs []string
```
mennopruijssers: Would be awesome if you could initialize `measureStrs` here with an initial capacity of `len(data.Measure)`. (not necessary but clearer and better performant)

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+		Desc:    desc,
+		Size:    size,
+		Script:  script,
+		Measure: measure,
+	}
+}
+
+func extractScript(script []map[string]string, varsData, runData []string) (labels, cmds []string) {
+	for _, cmdData := range script {
+		if len(cmdData) != 1 {
```
mennopruijssers: If you're expecting exactly 1 command, why is script an array/slice of map's instead of a string array/slice?

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+		Name:    name,
+		Desc:    desc,
+		Size:    size,
+		Script:  script,
+		Measure: measure,
+	}
+}
+
+func extractScript(script []map[string]string, varsData, runData []string) (labels, cmds []string) {
+	for _, cmdData := range script {
```
mennopruijssers: Would be awesome if you could initialize `labels` and `cmds` here with an initial capacity of len(script). (not necessary but clearer and better performant)

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: That's right, this'll have to change later but since the config of the testYaml is not yet declared. (I think it's pretty much defined in Sever and my head though) I will leave like this. Until Sever's cluster-manager is done, which is when we can actually be sure what we want to do with the config.

CorgiMan: Apologies for not indicating this earlier

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: Yes, this first line declares the variable names. A code comment is needed to indicate this.

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+
+	// extract size
+	size, err := strconv.Atoi(sizeStr)
+	if err != nil {
+		panic("size convert: " + err.Error())
+	}
+
+	// extract script
+	labels, cmds := extractScript(data.Script, varsData, runData)
+	script := parseScript(labels, cmds)
```
mennopruijssers: Why is is extracting and parsing scripts done in two functions instead of in one?

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	return &Command{
+		Label: label,
+		Cmd:   split[0],
+		Args:  split[1:],
+	}
+}
+
+func parseMeasure(msData []string) []*Measurement {
```
mennopruijssers: Naming: this should be parseMeasurements right?

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: No, when defining the runs you would say
runs:
- [<N>, <A>, <B>]

The < and > are the special characters. I'm currently not checking for this, do you think I should?

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+	return ms
+}
+
+func parseMeasurement(str string) *Measurement {
+	defer wrapPanicf("in parse measure '%s'", str)
+
+	split := split(str)
+	var measurementArgs = split
+
+	if len(split) < 3 {
```
mennopruijssers: after reading the code on lines 203-205, `!= 3` makes more sense to me?

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: I think it looks better when you specify a script in the yaml if it looks like:
```
- t0: kill 1
- t1: start 1
```

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+		panic("contains too few fields")
+	}
+
+	// search for assertion
+	var assertion *Assertion
+	for i, s := range split {
```
mennopruijssers: Shouldn't you ignore the first 3 items in `split`?
Or maybe even do something like this:
```
start := split[0]
end := split[1]
quantity := split[2]

// ignore first 3 parts
split := split[3:]
```

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+		if s == "is" || s == "in" {
+			interval := strings.Join(split[i+1:], "")
+			assertion = parseAssertion(split[i], interval)
+			measurementArgs = split[3:i]
```
mennopruijssers: Shouldn't you break out of the loop here?

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+		Quantity:  split[2],
+		Args:      measurementArgs,
+		Assertion: assertion,
+	}
+}
+
+func parseAssertion(typStr string, arg string) *Assertion {
+	defer wrapPanicf("in parse assertion '%s %s'", typStr, arg)
+
+	switch typStr {
```
mennopruijssers: `typeStr` makes more sense?

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+	for _, mData := range msData {
+		ms = append(ms, parseMeasurement(mData))
+	}
+	return ms
+}
+
+func parseMeasurement(str string) *Measurement {
+	defer wrapPanicf("in parse measure '%s'", str)
+
+	split := split(str)
```
mennopruijssers: `splitted` makes more sense?

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: will do

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+	}
+
+	panic("not valid assertion type")
+}
+
+func parseRange(rng string) (v1, v2 Value) {
+	defer wrapPanicf("in parse range '%s'", rng)
+
+	if rng[0] != '(' || rng[len(rng)-1] != ')' {
```
mennopruijssers: Will panic with out of range if `rng` is an empty string

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+		panic(fmt.Sprintf("types %T %T should be equal", v1, v2))
+	}
+
+	return v1, v2
+}
+
+func parseValue(str string) Value {
+	defer wrapPanicf("in parse value '%s", str)
+
+	// First check if the input is a numer or expression.
```
mennopruijssers: typo: `numer` -> `number`

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+// split splits a whitespace separted string
+func split(str string) []string {
+	defer wrapPanicf("in split '%s'", str)
+
+	scanner := bufio.NewScanner(strings.NewReader(str))
+	scanner.Split(bufio.ScanWords)
+
+	var wrds []string
```
mennopruijssers: Isn't it way easier to use `strings.Fields` (https://golang.org/pkg/strings/#Fields)?

cluster-test/test-orchestrator/test_yaml_parser.go
mennopruijssers: So we're misusing yaml for readability reasons? I'm ok with that but it might be useful to state it clearly when we're documenting the yaml-format

cluster-test/test-orchestrator/test_yaml_parser.go
mennopruijssers: Yeah, adding the check when parsing the vars might be nice. otherwise if I define a var called `i` all i's in all words will be replaced (E.g. a description like "check if ringpop reincarnates nodes" becomes "check 5f r5ngpop re5ncarnates nodes" when i=5).

mennopruijssers:
Adding unit-tests for some of these more generic methods might make sense as well

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: Because the functions where bigger previously it got a bit unwieldy having them in one function. I think it still makes sense to have them split, maybe not as much as previously though.

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: Yep.

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: Not really, it's parsing the measure field in the yaml. Have no problem changing it though

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: We could doesn't really matter to me.

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: We want the last is or in, just in case there is ever an in or is as an argument. It shouldn't matter though.

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: Ok

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: disagree but I'll call it words

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: That's okay since we use recover

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: It is, I did not know that existed. 
Changing...

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: We don't know the size beforehand. Its the sum of the number of runs for every scenario.

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: Ok

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: You commented the line so it seems like this isn't resolved but it is called fields now.

CorgiMan:
I think adding tests for this adds no value. For the other cases it added some insight but it's a lot of work to do every case. And if the parsing isn't correct you'll find that out immediately when you run the tests.

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -116,9 +118,9 @@ func extractScenario(data *scenarioYaml, runIx int) *Scenario {
 	script := parseScript(labels, cmds)
 
 	// extract Measure
-	var measureStrs []string
-	for _, measureStr := range data.Measure {
-		measureStrs = append(measureStrs, replace(measureStr, varsData, runData))
+	measureStrs := make([]string, len(data.Measure))
```
mennopruijssers: The other option here is:
```
var mesuareStrs := make([]string, 0, len(data.Measure))
for _m measureStr := range.Measure {
    measureStrs = append(measureStrs, replace(measureStr, varsData, runData))
}
```

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -134,8 +136,13 @@ func extractScenario(data *scenarioYaml, runIx int) *Scenario {
...
 		if len(cmdData) != 1 {
-			msg := fmt.Sprintf("\"%v\" is not a valid command, should contain exactly one entry", cmdData)
-			panic(msg)
+			// We are asserting that commands are one line only to comply with
+			// the yaml that looks like:
+			//
+			// script:
+			// - t0: command1
+			// - t1: command2
+			panic("command should contain exactly one entry")
```
mennopruijssers: We lost the command in the error message. Might be easier to add it again?

cluster-test/test-orchestrator/test_yaml_parser.go
mennopruijssers: Because you expect `len(cmdData)` to be `1` and append once per iteration of `script` the length of `labels` and `cmds` is always `len(script)` right?

cluster-test/test-orchestrator/test_yaml_parser.go
mennopruijssers: But the panic will say 'out of range' instead of saying something like 'missing range`

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: This check is added now

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: Because of the wrapPanicf, the line above it has `
```
in parse measurement 't0 t1 kill 1 in'
in parse assertion 'in '
in parse range ''
```
So I think it's clear enough then where the error lies. Do you agree?

cluster-test/test-orchestrator/test_yaml_parser.go
CorgiMan: Ah you're right, was thinking about runs vs scenarios extraction...

cluster-test/test-orchestrator/test_yaml_parser.go
mennopruijssers: Sorry you're right, I thought this one was missing the `wrapPanicf`

cluster-test/test-orchestrator/test_yaml_parser.go
```diff
@@ -0,0 +1,307 @@
...
+	return extractScenarios(testYaml)
+}
+
+// extractScenarios returns a scenario for every element in the runs list.
+func extractScenarios(runs *testYaml) []*Scenario {
+	var result []*Scenario
+	for _, scenarioData := range runs.Scenarios {
+		for _, vari := range scenarioData.Runs[0] {
+			if vari[0] != '<' || vari[len(vari)-1] != '>' {
+				panic(fmt.Sprintf("variable '%s' not of the from <var>", vari))
```
mennopruijssers: typo: from -> form

mennopruijssers:
one typo but besides that: LGTM

PR #82, opened 2016-06-27
Improve running_ringpop.md

Motiejus:
* Add note about broken table in RST
* change `ring.change` stat from gauge to counter
* fix documented location of tick-cluster
* add {membership,ring}.checksum-periodic to the stat table.

Related: https://github.com/uber/ringpop-go/pull/156


mennopruijssers:
It is possible to use inline html in markdown (https://daringfireball.net/projects/markdown/syntax#html).
Is that an option?
The other option is to convert this file to RST format and use the RST-table format right?

I think I would prefer one of those two options over a comment linking to the github page...

PR #83, opened 2016-07-22
bump tchannel in integration tests

mennopruijssers:
Current tchannel is outdated. By bumping it, the tests become compatible with node v4. (due to updated sse4_crc32.)


mennopruijssers:
Closing; tchannel is already udpated

PR #84, opened 2016-08-09
Sigterm from tick-cluster

mennopruijssers:
To assist in testing self eviction add sending `SIGTERM` to kill running processes.
Note: dependent on `SIGTERM` handling in testpop (see uber/ringpop-node#301)

### DIFF
--- a/tools/tick-cluster.js
+++ b/tools/tick-cluster.js
@@ -559,7 +559,7 @@ function displayMenu(logFn) {
     logFn('\th\t\tHelp menu');
     logFn('\tj\t\tJoin nodes');
     logFn('\tk <count>\tKill processes');
-    logFn('\tm <count>\Terminae processes');
+    logFn('\tm <count>\tTerminate processes');
     logFn('\tK\t\tRevive suspended or killed processes');
     logFn('\tl <count>\tSuspend processes');
     logFn('\tp\t\tPrint out protocol stats');
--- a/tools/tick-cluster.js
+++ b/tools/tick-cluster.js
@@ -505,9 +505,17 @@ function terminateProc(count) {
         .value();
 
     _.each(processesToKill, function kill(proc) {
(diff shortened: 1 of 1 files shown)


tools/tick-cluster.js
```diff
@@ -493,6 +498,19 @@ function killAllProcs() {
...
+function terminateProc(count) {
+    var processesToKill = _.chain(procs)
+        .filter(function (proc) { return !proc.killed && !proc.suspended; })
+        .sampleSize(+count)
+        .value();
+
+    _.each(processesToKill, function kill(proc) {
+        logMsg(proc.port, color.green('pid ' + proc.pid) + color.red(' randomly selected for termination'));
+        process.kill(proc.proc.pid, 'SIGTERM');
+        proc.killed = Date.now();
```
thanodnl: I think we need to set a timeout a couple of seconds to verify that the application has been shutdown because an application can consume the `SIGTERM` without actually quitting. If the application did not stop after that timeout is should be send `SIGKILL`.

mennopruijssers: Shouldn't that be a manual step of the tick-cluster user (`k`)?

thanodnl:
LGTM

tools/tick-cluster.js
btromanova: I agree with Nils. Probably we should put process to killed only when it's actually killed.

tools/tick-cluster.js
```diff
@@ -541,6 +567,7 @@ function displayMenu(logFn) {
     logFn('\th\t\tHelp menu');
     logFn('\tj\t\tJoin nodes');
     logFn('\tk <count>\tKill processes');
+    logFn('\tm <count>\tTerminate processes');
```
thanodnl: can this be moved down a bit to keep the letters in alphabetical order

btromanova:
lgtm

PR #85, opened 2016-08-24
Add integration test for self-eviction

mennopruijssers:
This test validates self-eviction by:
- stop gossiping in the whole cluster
- send a `SIGTERM ` to SUT
- validate if the SUT declares itself as faulty by pinging other members with the change

### DIFF
--- a/test/self-eviction-tests.js
+++ b/test/self-eviction-tests.js
@@ -60,8 +60,12 @@ function assertValidPings(t, tc, n) {
             type: events.Types.Ping,
             direction: 'request'
         });
+
         if (pings.length === 0) {
-            return cb(null);
+            // We should have at least one ping here,
+            // otherwise self-eviction is not working.
+            t.fail('no pings received');
+            return cb(pings);
         }
 
         var maxPings = Math.ceil(n * 0.4);
--- a/test/self-eviction-tests.js
+++ b/test/self-eviction-tests.js
@@ -51,6 +51,7 @@ test2('self eviction changes state to faulty and pings members', getClusterSizes
  *
  * @param t the current test suite
  * @param tc the test coordinated
+ * @param n the number of servers in the cluster
  * @return {assertValidPings} the assert functions
  */
 function assertValidPings(t, tc, n) {
--- a/test/self-eviction-tests.js
+++ b/test/self-eviction-tests.js
@@ -1,0 +1,82 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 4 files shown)


test/test-coordinator.js
```diff
@@ -303,7 +303,7 @@ TestCoordinator.prototype.shutdown = function shutdown() {
     if(this.adminChannel.topChannel.destroyed === false) {
         this.adminChannel.topChannel.close();
     }
-    this.sutProc.kill();
+    this.sutProc.kill('SIGKILL');
```
mennopruijssers: This change is necessary to force kill the the sut when the test is completed.

test/self-eviction-tests.js
```diff
@@ -0,0 +1,82 @@
...
+var test2 = require('./test-util').test2;
+
+
+test2('self eviction changes state to faulty and pings members', getClusterSizes(2), 2000000, prepareCluster(function(t, tc, n) {
+    return [
+        // disable pinging from fake nodes
+        dsl.disableAllNodesPing(t, tc),
+
+        // stop gossiping on SUT
+        dsl.stopGossip(t, tc),
```
btromanova: Can you pls add a comment to justify stopGossip and disableAllNodesPing?

test/self-eviction-tests.js
```diff
@@ -0,0 +1,82 @@
...
+        var pings = _.filter(list, {
+            type: events.Types.Ping,
+            direction: 'request'
+        });
+        if (pings.length === 0) {
+            return cb(list);
+        }
+
+        var maxPings = Math.ceil(n * 0.4);
+        t.ok(pings.length <= maxPings, 'number of pings does not exceed 40% of cluster');
```
btromanova: Do we need also to check that at least some number of pings are sent?

mennopruijssers: Good catch. I changed line 62 to `return cb(null)`. That'll make sure the assertValidPings method is rerun if there are 0 pings.

btromanova:
lgtm

test/self-eviction-tests.js
```diff
@@ -0,0 +1,83 @@
...
+}));
+
+/**
+ * Validate the pings. Assert that:
+ * - the number of pings doesn't exceed 40% of the cluster
+ * - the pings are all to different members
+ * - the pings should declare the SUT faulty and originate from the SUT itself
+ *
+ * @param t the current test suite
+ * @param tc the test coordinated
```
thanodnl: the param for `n` is missing.

test/self-eviction-tests.js
```diff
@@ -0,0 +1,83 @@
...
+ * @return {assertValidPings} the assert functions
+ */
+function assertValidPings(t, tc, n) {
+    return function assertValidPings(list, cb) {
+        var pings = _.filter(list, {
+            type: events.Types.Ping,
+            direction: 'request'
+        });
+        if (pings.length === 0) {
```
thanodnl: If I remember correctly this assertion function will be called for every incoming packet directly. This branch will not be hit as soon as the first faulty declaration is received. causing all subsequent  assertions to only be done on the first ping.

I think we should change the `pings.length === 0` to `pings.length !== expectedFaultyPings` to make sure we assert on the correct set of pings.

mennopruijssers: It's called after `waitForGracefulShutdown` (which waits until the process is exited).

PR #86, opened 2016-08-24
Add integration tests for ringpop-labels

thanodnl:
This PR adds tests for the labels feature in ringpop to the integration test suit. Currently the feature is only implemented in ringpop-go and the test is a non mandatory test for now and enabled via the `--enable-feature labels` command line option.

### DIFF
--- a/schema/change.json
+++ b/schema/change.json
@@ -28,12 +28,7 @@
             "type": "boolean"
         },
         "labels": {
-            "type": "object",
-            "patternProperties": {
-                "^.*$": {
-                    "type": "string"
-                }
-            }
+            "$ref": "/Labels"
         }
     },
     "required": [
--- a/schema/change.json
+++ b/schema/change.json
@@ -26,6 +26,14 @@
         },
         "tombstone": {
             "type": "boolean"
+        },
+        "labels": {
+            "type": "object",
+            "patternProperties": {
+                "^.*$": {
+                    "type": "string"
+                }
+            }
         }
     },
     "required": [
--- a/schema/labels.json
+++ b/schema/labels.json
@@ -1,0 +1,10 @@
+{
+    "id": "/Labels",
+    "title": "Labels",
+    "type": "object",
+    "patternProperties": {
+        "^.*$": {
+            "type": "string"
+        }
+    }
+}
--- a/test/fake-node.js
+++ b/test/fake-node.js
@@ -111,12 +111,22 @@ FakeNode.prototype.shutdown = function shutdown() {
 };
 
 FakeNode.prototype.toMemberInfo = function toMemberInfo() {
-    return {
+    var memberInfo = {
         host: this.host,
         port: this.port,
         status: this.status,
         incarnationNumber: this.incarnationNumber,
     };
+
+    if (this.labels) {
+        // the easiest way to get the labels in the join response.
+        if (!memberInfo.extraFields) {
+            memberInfo.extraFields = {};
+        }
+        memberInfo.extraFields.labels = this.labels;
+    }
+
+    return memberInfo;
 };
 
 FakeNode.prototype.changeEndpoint = function modifyEndpoint(endpoint, handler) {
@@ -127,6 +137,8 @@ FakeNode.prototype.changeEndpoint = function modifyEndpoint(endpoint, handler) {
 // Useful, e.g., for partitioning tests.
 FakeNode.prototype.joinHandler = function joinHandler(req, res, arg2, arg3) {
     var membership = this.membership || this.coordinator.getMembership();
+
+    console.log("membership:",JSON.stringify(membership, null, 2));
     return handleJoin(req, res, this.toMemberInfo(), membership, this.coordinator.checksum);
 };
 
--- a/test/fake-node.js
+++ b/test/fake-node.js
@@ -137,8 +137,6 @@ FakeNode.prototype.changeEndpoint = function modifyEndpoint(endpoint, handler) {
 // Useful, e.g., for partitioning tests.
 FakeNode.prototype.joinHandler = function joinHandler(req, res, arg2, arg3) {
     var membership = this.membership || this.coordinator.getMembership();
-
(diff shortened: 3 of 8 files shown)


test/fake-node.js
```diff
@@ -127,6 +137,8 @@ FakeNode.prototype.changeEndpoint = function modifyEndpoint(endpoint, handler) {
 // Useful, e.g., for partitioning tests.
 FakeNode.prototype.joinHandler = function joinHandler(req, res, arg2, arg3) {
     var membership = this.membership || this.coordinator.getMembership();
+
+    console.log("membership:",JSON.stringify(membership, null, 2));
```
thanodnl: need to remove this log line.

schema/change.json
```diff
@@ -26,6 +26,14 @@
         },
         "tombstone": {
             "type": "boolean"
+        },
+        "labels": {
```
mennopruijssers: It might be nice to extract the next couple of lines in it's own object and re-use it from `admin-stats-response.json`

test/labels.js
```diff
@@ -0,0 +1,349 @@
...
+                1: { labels: labels2 }
+            }),
+        ];
+    })
+);
+
+// since nodes might miss updates on changes they should always accept labels if
+// the status updates, even if these labels differ from the labels that were
+// previously known to the node.
+function testLabelOverrideOnStatusChange(firstStatus, secondStatus) {
```
mennopruijssers: I would prefer if this function is nearby the callers (starting on line 334)

test/membership-checksum.js
```diff
@@ -31,9 +31,47 @@ function generateChecksumString(members) {
...
             var address = member.address || (member.host + ':' + member.port);
-            return address +
-                    member.status +
-                    member.incarnationNumber
+
+            var labelChecksum = 0;
+
+            if (member.labels) {
+                Object.keys(member.labels).forEach(function (key) {
+                    var keyBuffer = new Buffer(key);
```
mennopruijssers: nonblocking but you could just create one buffer and append on it

mennopruijssers:
Looks really good! Three non-blocking nits (and the `console.log` you already found)

test/labels.js
thanodnl: I agree, for readability the invocations will be first and the definition later.

test/membership-checksum.js
thanodnl: buffers in nodejs can't be resized after initialization. Since the size of the required buffer is unknown due to the encoding of strings this was the easiest way of calculating the final buffer.

schema/change.json
thanodnl: The definition of labels has been extracted.

test/membership-checksum.js
mennopruijssers: `var bufferLength = 4 + Buffer.byteLength(key) + 4 + Buffer. byteLength(member.labels[key])`?

schema/labels.json
```diff
@@ -0,0 +1,10 @@
+{
+    "id": "/Labels",
+    "title": "Labels",
+    "type": "object",
+    "patternProperties": {
+        "^.*$": {
```
mennopruijssers: 👍 !

test/membership-checksum.js
```diff
@@ -31,9 +31,45 @@ function generateChecksumString(members) {
...
+                    var pos = 0;
+                    var buf = new Buffer(8 + keyLength + valueLength);
+                    buf.writeInt32BE(keyLength, pos);
+                    pos += 4
+                    pos += buf.write(key, pos);
+                    buf.writeInt32BE(valueLength, pos);
+                    pos += 4
+                    pos += buf.write(value, pos);
+
+                    labelChecksum ^= farmhash.fingerprint32(buf);
```
mennopruijssers: ❤️

mennopruijssers:
LGTM!

test/labels.js
```diff
@@ -0,0 +1,349 @@
...
+        dsl.assertStats(t, tc, n+1, 0, 0, {
+            1: {
+                labels: undefined
+            }
+        }),
+    ];})
+);
+
+// The most likely assert to fail will be the last assert in this test where the
+// winning labels are compaired to the labels that were chosen after the first
```
dansimau: s/compaired/compared

test/labels.js
```diff
@@ -0,0 +1,349 @@
...
+// round of label gossipping
+test2('ringpop should deterministically pick labels on conflict', getClusterSizes(2), 20000,
+    prepareCluster(function(t, tc, n) {
+        // keep track of the labels that won the first time around
+        var winningLabels = {};
+
+        // two pairs of labels
+        var labels1 = { "hello": "world" };
+        var labels2 = { "hello": "goodbey" };
```
dansimau: Is `goodbey` a typo, or did you mean, like, you're telling your significant other that he/she has been good?

beyonce wink

test/labels.js
```diff
@@ -0,0 +1,349 @@
...
+            // as the first iteration to ensure convergence when there are
+            // conflicting labels being gossipped around.
+            dsl.assertMembership(t, tc, {
+                1: {
+                    labels: winningLabels
+                }
+            }),
+        ];
+    })
+);
```
dansimau: Nice one

dansimau:
Looks pretty awesome. Just one question: are we testing upgrade/rollback behaviours in here?

thanodnl:
The integration tests don't really test the upgrade/rollback behavior of ringpop. It is a great idea but I am not sure that it can be easily tested. Most often what actually happens goes into your gut feeling so I think it is best to keep testing those paths with a test cluster.

When I ran an upgrade without using labels it worked as always. No faulties, no fullsyncs. When having labels set during an upgrade or rollback there is slightly more gossip due to continues reincarnation of nodes with labels. The old version will treat the member as if it has no labels and gossips that around, this will reach the member in question which disagrees and reincarnates.

If you have ideas on how to test these scenarios it would be great to have a chat about it.

dansimau:
Sweet, LGTM

PR #87, opened 2016-08-26
Validate schema of events after test is complete

mennopruijssers:
Before this PR, we were validating each event when it happens. Especially on slower machines (e.g. travis), the impact is severe enough to interfere with some tests. Local testing showed that (with cpu throttled to 5%), event validation takes over 2.5 seconds for the bidirectional full sync tests.

### DIFF
--- a/test/ringpop-assert.js
+++ b/test/ringpop-assert.js
@@ -705,11 +705,10 @@ function enableNode(t, tc, ix, incarnationNumber) {
     return f;
 }
 
-function createValidateEvent(t, tc) {
-    var Validator = jsonschema.Validator;
-    var validator = new Validator();
+function createEventValidator(t) {
+    var validator = new jsonschema.Validator();
 
-    // load all json schema files and add them to the valicator
+    // load all json schema files and add them to the validator
     var schemaFiles = glob.sync("../schema/*.json",{cwd:__dirname});
     schemaFiles.forEach(function (schemaFile) {
         validator.addSchema(require(schemaFile));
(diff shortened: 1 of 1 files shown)


test/ringpop-assert.js
```diff
@@ -791,11 +790,10 @@ function createValidateEvent(t, tc) {
 function validate(t, tc, scheme, deadline) {
     var fns = scheme;
     var cursor = 0;
+    var allEvents = [];
     var eventList = [];
 
-    tc.on('event', createValidateEvent(t, tc));
```
mennopruijssers: Validation is moved to line 866

test/ringpop-assert.js
```diff
@@ -824,12 +822,7 @@ function validate(t, tc, scheme, deadline) {
         inProgress = true;
 
         if(cursor >= fns.length) {
-            clearTimeout(timer);
-            t.ok(true, 'validate done: all functions passed');
-            tc.shutdown();
-            tc.removeAllListeners('event');
-            tc.removeAllListeners('sut-died');
-            t.end();
+            progressionComplete();
```
mennopruijssers: These lines are extracted in the `progressionComplete`-function.

test/ringpop-assert.js
```diff
@@ -858,8 +851,23 @@ function validate(t, tc, scheme, deadline) {
 
     tc.on('event', function(event) {
         eventList.push(event);
+        allEvents.push(event);
```
mennopruijssers: We need a second list, because events that are handled by the test are removed from the `eventList`.

severb:
LGTM!

PR #88, opened 2016-09-13
Fix flappy integration test

mennopruijssers:
This PR fixes three issues:
- [x] A race condition where a test completes and times out at the same time: this is fixed by breaking out the event-loop early after a time out.
- [x] Start running the test before the SUT is ready: fixed by waiting for an initial ping from the SUT.
- [x] The implementation of lookup in the integration tests is incorrect. Sometimes it returns an incorrect result

### DIFF
--- a/test/test-coordinator.js
+++ b/test/test-coordinator.js
@@ -102,36 +102,34 @@ TestCoordinator.prototype.lookup = function(key) {
     });
     hostPorts.push(this.sutHostPort);
 
-    // iterate over all hosts
-    var allHashes = [];
-    hostPorts.forEach(function (hostPort) {
+    var minimumHash, minimumNode;
+    var bestHash, bestNode;
+
+    hostPorts.forEach(function (hostPort){
         // iterate over all vnodes for that host
         for (var i=0; i<self.replicaPoints; i++) {
-            var currentHash = farmhash.fingerprint32(hostPort + i);
-            allHashes.push({node: hostPort, hash: currentHash});
-        }
-    });
+            var vNodeHash = farmhash.fingerprint32(hostPort + i);
 
-    allHashes = allHashes.sort(function(a, b) {
-        return a.hash - b.hash;
-    });
+            if (minimumHash === undefined || vNodeHash < minimumHash) {
+                minimumHash = vNodeHash;
+                minimumNode = hostPort;
+            }
 
-    var nodeHash;
-
-    if (allHashes[allHashes.length - 1].hash < hash) {
-        // If the last hash is smaller than the hash of the key we need to wrap around.
-        nodeHash = allHashes[0];
-    } else {
-
-        for (var i = 0; i < allHashes.length; i++) {
-            if (allHashes[i].hash >= hash) {
-                nodeHash = allHashes[i];
-                break;
+            // If the vnodehash is after the hash and before the current best, select it.
+            if (vNodeHash >= hash && (bestHash === undefined || vNodeHash < bestHash)) {
+                bestHash = vNodeHash;
+                bestNode = hostPort;
             }
         }
-    }
+    });
 
-    return nodeHash.node;
+    // All vNodes are below the key: wrap around
+    if (bestNode === undefined) {
+        return minimumNode;
+    }
+    else {
+        return bestNode;
+    }
 };
 
 TestCoordinator.prototype.createFakeNode = function createFakeNode() {
@@ -327,7 +325,6 @@ TestCoordinator.prototype.createHostsFile = function createHostsFile(hostPortLis
     if (!hostPortList) {
         hostPortList = this.getStandardHostPortList();
     }
-    console.log('hosts: ', hostPortList);
     fs.writeFileSync(this.hostsFile, JSON.stringify(hostPortList));
 };
 
--- a/test/test-coordinator.js
+++ b/test/test-coordinator.js
@@ -95,21 +95,7 @@ require('util').inherits(TestCoordinator, events.EventEmitter);
 
 TestCoordinator.prototype.lookup = function(key) {
     var self = this;
-
-    var matchingNode;
-    var matchingHash;
-    var minimumNode;
-    var minimumHash;
-
     var hash = farmhash.fingerprint32(key);
-
-    function isBetterMatch(newHash) {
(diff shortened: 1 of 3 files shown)


mennopruijssers:
Test's pass locally on both go and node:

ringpop-go:
```
make test-integration
test/run-integration-tests
[fetch ringpop-common] + cd test/ringpop-common
[fetch ringpop-common] + git fetch origin fix-flappy
[build testpop] + make testpop
[build testpop] rm -f testpop
[build testpop] go build ./scripts/testpop/
[fetch ringpop-common] From ssh://github.com/uber/ringpop-common
[fetch ringpop-common]  * branch            fix-flappy -> FETCH_HEAD
[fetch ringpop-common] + git checkout FETCH_HEAD
[fetch ringpop-common] Previous HEAD position was 4e35f91... wait for first ping in prepare cluster
[fetch ringpop-common] HEAD is now at 0209af7... Don’t conumse initial ping
[fetch ringpop-common] + cd -
[fetch ringpop-common] + cd test/ringpop-common/test
[fetch ringpop-common] + cd -
[test-runner] Running test for cluster size 1...
[test-runner] Running test for cluster size 2...
[test-runner] Running test for cluster size 3...
[test-runner] Running test for cluster size 4...
[test-runner] Running test for cluster size 5...
[test-runner] Running test for cluster size 10...
Tests passed
```

ringpop-node
```
npm run test-shared-integration-tests

> ringpop@10.15.0 test-shared-integration-tests /Users/menno/Sources/Uber/nodejs/ringpop-node
> test/run-shared-integration-tests

[fetch ringpop-common] + cd test/ringpop-common
[fetch ringpop-common] + git fetch origin fix-flappy
[fetch ringpop-common] From ssh://github.com/uber/ringpop-common
[fetch ringpop-common]  * branch            fix-flappy -> FETCH_HEAD
[fetch ringpop-common] + git checkout FETCH_HEAD
[fetch ringpop-common] Previous HEAD position was 4e35f91... wait for first ping in prepare cluster
[fetch ringpop-common] HEAD is now at 0209af7... Don’t conumse initial ping
[fetch ringpop-common] + cd -
[fetch ringpop-common] + cd test/ringpop-common/test
[fetch ringpop-common] + npm install
[fetch ringpop-common] npm WARN package.json ringpop-common@0.0.1 No license field.
[fetch ringpop-common] + cd -
[test-runner] Running test for cluster size 1...
[test-runner] Running test for cluster size 2...
[test-runner] Running test for cluster size 3...
[test-runner] Running test for cluster size 4...
[test-runner] Running test for cluster size 5...
[test-runner] Running test for cluster size 10...
Tests passed
```

test/test-util.js
```diff
@@ -121,8 +121,9 @@ function testStateTransitions(ns, initial, newState, finalState, incNoDelta, sta
 function prepareCluster(insert_fns) {
     return function(t, tc, n) {
         return [
-            dsl.assertDetectChecksumMethod(t, tc),
+            dsl.waitForPing(t, tc, false),
```
thanodnl: Maybe we should describe why we are waiting for a ping, and why we don't consume it. Both are very subtile but important facts.

thanodnl:
LGTM

thanodnl:
Nice find on the bug in the lookup function of the test coordinator, fix is legit.

test/test-util.js
```diff
@@ -121,8 +121,13 @@ function testStateTransitions(ns, initial, newState, finalState, incNoDelta, sta
 function prepareCluster(insert_fns) {
     return function(t, tc, n) {
         return [
-            dsl.assertDetectChecksumMethod(t, tc),
+            // By waiting for the first ping we make sure the SUT is ready to go.
+            // We don't consume it so other tests (especially the ping-tests)
+            // can still assert implementation details related to ping distribution.
+            dsl.waitForPing(t, tc, false),
```
CorgiMan: I'm a bit confused. Shouldn't the joins happen before the joins?

test/test-coordinator.js
```diff
@@ -95,42 +95,43 @@ require('util').inherits(TestCoordinator, events.EventEmitter);
...
-        return false;
-    }
-
     var hostPorts = this.fakeNodes.map(function (fn) {
         return fn.getHostPort();
     });
     hostPorts.push(this.sutHostPort);
 
     // iterate over all hosts
+    var allHashes = [];
```
CorgiMan: I found this hard to follow but maybe I'm just looking at it differently.

Can't we just do something like:
```
var min = Infinity;
var node = undefined;

hostPorts.forEach(function (hostPort) {
  for (var i=0; i<self.replicaPoints; i++) {
    var replica = farmhash.fingerprint32(hostPort + i);
    var dist = Math.min(replica - hash, replica-hash + Math.pow(2, 32))
    if (dist < min) {
        min = dist;
        node = hostPort;
    }
  }
}
```

Again, I could just be looking at it differently miss a subtlety..

test/test-util.js
mennopruijssers: The idea behind it is to wait until the node starts pinging. At that point we can consider it to be fully ready and start running the other validations.

test/test-coordinator.js
mennopruijssers: I agree that it's pretty hard to follow but I'm not sure if you're version is the most readable option. I'll try to restructure this method again to try to improve it even further.

test/test-util.js
CorgiMan: But wouldn't it make sense to first wait for the joins, and then for the pinging. I don't think the node will ping before it has joined.

test/test-coordinator.js
CorgiMan: Some comment's will also be sufficient.

test/test-util.js
mennopruijssers: Makes sense, I'll reorder them

mennopruijssers: After a discussion with Nils we changed it back (and added a comment). The reason why we're waiting for the initial ping first is so we wait until the node is fully ready before checking the number of joins. This way we're able to detect if there are too many joins during standup.

CorgiMan: makes sense, thanks!

CorgiMan:
Looks good!

PR #89, opened 2016-12-06
Allow multiple checksums

mennopruijssers:
uber/ringpop-go#192 adds support for multiple checksums. This PR makes the required changes to the protocol schema.

### DIFF
--- a/schema/admin-stats-response.json
+++ b/schema/admin-stats-response.json
@@ -115,6 +115,14 @@
                 "checksum": {
                     "type": "number"
                 },
+                "checksums": {
+                    "type": "object",
+                    "patternProperties": {
+                        "^.*$": {
+                            "type": "number"
+                        }
+                    }
+                },
                 "servers": {
                     "type": "array",
                     "items": {


schema/admin-stats-response.json
```diff
@@ -115,6 +115,14 @@
                 "checksum": {
                     "type": "number"
                 },
+                "checksums": {
```
mennopruijssers: Since ringpop-node doesn't support it yet, `checksums` is not added to the list with required properties.

dansimau: 👍

schema/admin-stats-response.json
```diff
@@ -115,6 +115,14 @@
...
+                    "type": "object",
+                    "patternProperties": {
+                        "^.*$": {
+                            "type": "number"
```
dansimau: A checksum is always a number?

mennopruijssers: For now yes (and I don't see any reason to change it in the future)

dansimau: The result of a `farmhash` is always a number. That's the bit I was missing (I thought it was alphanumeric). Cool, thanks!

PR #90, opened 2016-12-06
Add integration tests for lookups

mennopruijssers:
Integration tests for `lookups` were missing. Since we're touching it for identity carryover, the right thing to do is add them so we're sure we don't break backwards compatibility.


mennopruijssers:
> Do you think we should assert the result of lookups when a member has transitioned to other statuses too (e.g. suspect, etc)?

Makes sense, I'll add those as well.

PR #91, opened 2016-12-13
Allow initial labels

mennopruijssers:
A pre-requirement for identity-carryover is support to set labels before bootstrapping. 
To improve support for labels and extend this with labels set before bootstrapping, 
- Add labels to `join-request` and `join-response` schema.
- store and propagate labels from SUT and fake nodes correctly

### DIFF
--- a/test/ringpop-assert.js
+++ b/test/ringpop-assert.js
@@ -52,8 +52,10 @@ function waitForJoins(t, tc, n) {
         t.equals(joins.length, n, 'check number of joins',
             errDetails({journal: _.pluck(list, 'endpoint')}));
 
-        //XXX: a bit inappropriate to get sutIncarnationNumber like this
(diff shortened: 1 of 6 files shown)


test/ringpop-assert.js
```diff
@@ -52,8 +52,10 @@ function waitForJoins(t, tc, n) {
         t.equals(joins.length, n, 'check number of joins',
             errDetails({journal: _.pluck(list, 'endpoint')}));
 
-        //XXX: a bit inappropriate to get sutIncarnationNumber like this
-        tc.test_state['sutIncarnationNumber'] = safeJSONParse(joins[0].arg3).incarnationNumber;
+        //XXX: a bit inappropriate to set internal state like this
+        var joinRequest = safeJSONParse(joins[0].arg3);
```
thanodnl: I think you can use `.body` here as a parsed version of `arg3` already. It hasn't been changed throughout the code, but now that you are changing it you might as well use it.

PR #94, opened 2016-12-19
Fix instructions to require modules

dschenkelman:
Changed `require('TChannel')` to `require('tchannel')`. The module name is correct in the latter.


mennopruijssers:
@dschenkelman thanks for improving the docs! If you sign or [CLA](https://cla-assistant.io/uber/ringpop-common?pullRequest=94), I'll make sure this is merged to master. Thanks!

PR #95, opened 2016-12-20
Test state transitions

mennopruijssers:
Test if state transitions (`suspect` -> `faulty` and  `faulty` -> `tombstone`) are correctly scheduled when the initial membership contains a member with a `suspect` or `tombstone` status.

### DIFF
--- a/test/join-tests.js
+++ b/test/join-tests.js
@@ -63,3 +63,36 @@ test2('join ringpop with fake node', getClusterSizes(), 20000,
         dsl.expectOnlyPings(t, tc),
     ];})
 );
+
+test2('5-second suspect to faulty window on join',
+    getClusterSizes(2),
+    20000,
+    function init(t, tc, callback) {
+        tc.addMembershipInformation('192.0.2.100:1234', 'suspect', 127);
+        callback();
+    },
+    prepareCluster({faulty: 1}, function(t, tc, n) {
+        return [
+            dsl.assertStats(t, tc, {
+                alive: n+1,
+                suspect: 1,
+                faulty: 0,
+                tombstone: 0
+            }),
+            dsl.wait(4000),
+            dsl.assertStats(t, tc, {
+                alive: n+1,
+                suspect: 1,
+                faulty: 0,
+                tombstone: 0
+            }),
+            dsl.wait(1100),
+            dsl.assertStats(t, tc, {
+                alive: n+1,
+                suspect: 0,
+                faulty: 1,
+                tombstone: 0
+            })
+        ];
+    })
+);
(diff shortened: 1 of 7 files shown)


test/ringpop-assert.js
```diff
@@ -625,6 +625,59 @@ function assertStats(t, tc, statusCountsOralive, suspect, faulty, members) {
     return result;
 }
 
+function assertStateChange(t, tc, addressOrIndex, status, expectedDuration, allowedJitter) {
```
thanodnl: This function measures the time from the first join but is used in both tests (state in join request and state updated after the join completed). Might be better to make that configurable I think and configure it correctly in both tests. This might reduce flaky-ness of the jitter.

test/test-coordinator.js
```diff
@@ -156,6 +156,11 @@ TestCoordinator.prototype.start = function start(callback) {
     var self = this;
     callback = callback || function(){};
 
+    self.once('joined', function joined(){
+        // Set firstJoinDate to current Date if not already set.
+        self.firstJoinDate = self.firstJoinDate || new Date();
```
thanodnl: I think keeping track of the last join date is easy and desirable as well. iirc most of the ringpop logics start after the last join.

test/join-tests.js
```diff
@@ -63,3 +63,23 @@ test2('join ringpop with fake node', getClusterSizes(), 20000,
...
+test2('5-second suspect to faulty window on join',
+    getClusterSizes(2),
+    20000,
+    function init(t, tc, callback) {
+        tc.addMembershipInformation('192.0.2.100:1234', 'suspect', 127);
+        callback();
+    },
+    prepareCluster({suspect: 1}, function(t, tc, n) {
+        return [
+            dsl.assertStats(t, tc, {
```
thanodnl: this test is already included in `prepareCluster` don't think we need it here again

test/test-coordinator.js
```diff
@@ -156,9 +156,15 @@ TestCoordinator.prototype.start = function start(callback) {
     var self = this;
     callback = callback || function(){};
 
-    self.once('joined', function joined(){
+    self.on('joined', function joined(){
         // Set firstJoinDate to current Date if not already set.
-        self.firstJoinDate = self.firstJoinDate || new Date();
+        self.firstJoinTime = self.firstJoinTime || Date.now();
+        self.lastJoinTime = Date.now();
+        console.log('got joined: ', self.firstJoinTime, self.lastJoinTime);
```
thanodnl: I don't think we want to start logging from tests. If there is a good reason to always log this we can keep it, but I'd rather see that we put all data needed to debug a failure in the failure of a test.

test/ringpop-assert.js
```diff
@@ -666,11 +666,13 @@ function assertStateChange(t, tc, addressOrIndex, status, expectedDuration, allo
         });
 
         if (index < 0) {
+
```
thanodnl: nit: white line

PR #96, opened 2016-12-20
Support rolling restart in tick-cluster

mennopruijssers:


tools/tick-cluster.js
```diff
@@ -519,6 +525,60 @@ function terminateProc(count) {
     });
 }
 
+function restartProc(batchSize) {
+    if (batchSize === 0) {
+        return;
```
thanodnl: maybe add output to indicate it is backing out because of invalid input

tools/tick-cluster.js
```diff
@@ -570,6 +630,7 @@ function displayMenu(logFn) {
     logFn('\tK\t\tRevive suspended or killed processes');
     logFn('\tl <count>\tSuspend processes');
     logFn('\tm <count>\tTerminate processes');
+    logFn('\tr <batch size>\tRestart processes');
```
thanodnl: I think the list is in alphabetic order, except for this line. Lets move it down a bit.

tools/tick-cluster.js
mennopruijssers: At all other places, `tick-cluster` silently ignore incorrect input

tools/tick-cluster.js
mennopruijssers: Yeah, you're right; I put it in logical order but the rest if alphabetic. I'll reorder.

